PR #3, opened 2016-07-14
FSET-580 Rename to Fast Stream

ollyw:


henricook:
Looks like you missed a few spots @ollyw  (see build)

PR #4, opened 2016-07-19
Add 'service'  auth-provider parameter + 'ping' test generator operation

lukastymo:


ollyw:
@lukastymo this works for me locally, and looks fine. It is failing on Travis, but suspect that is an intermittent mongodb issue, not related to this story

PR #7, opened 2016-07-26
FSET-588: Enter personal details

lukastymo:


ollyw:
@lukastymo Is this blocked awaiting the refactor of status codes? If so, can you also block https://github.com/hmrc/fset-faststream-frontend/pull/14 please?

lukastymo:
@ollyw It's read to be merged, in order to keep the end-to-end flow we need to merge faststream-frontend and faststream in the same time. In my next commit I'm going to improve test coverage for these controllers

app/repositories/personaldetails/PersonalDetailsRepository.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  def find(appId: String): Future[PersonalDetails]
+}
+
+class PersonalDetailsMongoRepository(implicit mongo: () => DB)
+  extends ReactiveRepository[PersonalDetails, BSONObjectID]("application", mongo, PersonalDetails.personalDetailsFormat,
+    ReactiveMongoFormats.objectIdFormats) with PersonalDetailsRepository {
+  val PersonalDetailsCollection = "personal-details"
+
+  def update(applicationId: String, userId: String, personalDetails: PersonalDetails,
```
ollyw: Should we consider passing in the allowed beginning states? This would make it safer to transition. E.g. the candidate shouldn't go from Application Status of withdrawn -> personal details updated. It is possible with the current scheme with concurrent modifications (although is very slim chance of it happening in this case)

mgilgar:
Generally speaking it looks good.

it/testkit/CommonHttpWorkflows.scala
```diff
@@ -49,9 +49,7 @@ trait CommonHttpWorkflows { self: WireLevelHttpSpec =>
            |   },
            |  "postCode":"H0H 0H0",
            |  "mobilePhone":"071234567",
-           |  "recentSchool":"Super hero's academy",
-           |  "aLevel": $aLevelsAboveD,
-           |  "stemLevel": $aLevelsAboveCInStem
+           |  "recentSchool":"Super hero's academy"
```
ollyw: Overly pedantic grammar police notice: Did you mean "Superheroes' academy", "Superhero's academy" or "Superheroes academy"

lukastymo: Removed

app/repositories/personaldetails/PersonalDetailsRepository.scala
lukastymo: Good point. It looks like these guard states are prerequisites and they are not part of evaluation in services. Maybe we should put them hardcoded in repositories instead passing from the service?

```
BSONDocument("applicationId" -> applicationId, "userId" -> userId, "$or" -> BSONArray(BSONDocument("applicationStatus" -> CREATED), BSONDocument("applicationStatus" -> IN_PROGRESS))
```

PR #16, opened 2016-08-19
FSET-669: Choose my scheme preferences

Lansalo:
Update model as with 669 the candidate won't have the option to consider alternatives schemes anymore.


Lansalo:
The merge of this bit has to be done together with the corresponding change on the frontend

Lansalo:
This if part of a bunch of PR related to FSET-669.
see FSET-669 on:
fset-faststream-frontend
fset-faststream
fset-faststream-candidate-acceptance-tests

There is a dependency between the frontend and backend and the test should be merged at the end.

PR #18, opened 2016-08-24
FSET-589: Fast pass details

satishkamatam:


app/services/generaldetails/CandidateDetailsService.scala
```diff
@@ -48,22 +49,27 @@ trait CandidateDetailsService {
     val updatePersonalDetailsFut = pdRepository.update(applicationId, userId, personalDetails,
       sourceStatuses = List(CREATED, IN_PROGRESS), targetStatus = IN_PROGRESS)
     val contactDetailsFut = cdRepository.update(userId, contactDetails)
+    val fastPassDetailsFut = fpdRepository.update(applicationId, candidateDetails.fastPassDetails)
```
mgilgar: Do we really need another repository for fastPassDetails, isn't it another part of the personal details? We can save a call to mongo. For ContactDetails it make sense because is a separate collection

app/controllers/CandidateDetailsController.scala
```diff
@@ -16,7 +16,7 @@
 
 package controllers
 
-import model.Exceptions.{CannotUpdateContactDetails, CannotUpdateRecord, ContactDetailsNotFound, PersonalDetailsNotFound}
+import model.Exceptions._
 import model.command.UpdateGeneralDetails
```
mgilgar: I think UpdateGeneralDetails is a bit misleading, it is also used in the find method, I think a better name would be command.GeneralDetails or maybe command.PersonalDetails

app/repositories/FastPassDetailsRepository.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  def find(applicationId: String): Future[FastPassDetails]
+
+}
+
+class FastPassDetailsMongoRepository(implicit mongo: () => DB) extends
+  ReactiveRepository[FastPassDetails, BSONObjectID]("application", mongo, FastPassDetails.fastPassDetailsFormat,
+    ReactiveMongoFormats.objectIdFormats) with FastPassDetailsRepository {
+
+  override def update(applicationId: String, fastPassDetails: FastPassDetails): Future[Unit] = {
```
mgilgar: Are we going to update fastpass details independently from other personal details? It might be good to keep it but as I said in previous comment, maybe this is part of the personal details.

app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -209,8 +209,9 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService)(implic
       case Some(document) =>
         val applicationId = document.getAs[String]("applicationId").get
         val applicationStatus = document.getAs[String]("applicationStatus").get
+        val fastPassReceived = document.getAs[FastPassDetails]("fastpass-details").flatMap(_.fastPassReceived)
         findProgress(applicationId).map { (p: ProgressResponse) =>
-          ApplicationResponse(applicationId, applicationStatus, userId, p)
+          ApplicationResponse(applicationId, applicationStatus, userId, p, fastPassReceived)
```
mgilgar: I am thinking maybe fastPassReceived is a kind of status or progress? I don't like the idea of storing that outside applicationStatus or progressStatus. And I know we will have like that, maybe we have to think how to deal with it.

it/repositories/FastPassDetailsRepositorySpec.scala
```diff
@@ -0,0 +1,49 @@
...
+    "modify and find the fast pass details successfully" in {
+      val fastPassDetails = (for {
+        _ <- insert(BSONDocument("applicationId" -> AppId, "userId" -> UserId, "applicationStatus" -> CREATED))
+        _ <- repository.update(AppId, civilServant)
+        fpDetails <- repository.find(AppId)
+      } yield fpDetails).futureValue
+      fastPassDetails mustBe civilServant
+    }
+
+    "return exception when application does not exist" in {
```
mgilgar: We do not really know if it exists or not, we just got CannotUpdate exception

app/repositories/FastPassDetailsRepository.scala
satishkamatam: As discussed we will keep it like this for now.

app/repositories/application/GeneralApplicationRepository.scala
satishkamatam: As discussed at the moment its not relevant to progress. We may get an idea when we do the stories related to civil service.

it/repositories/FastPassDetailsRepositorySpec.scala
satishkamatam: wording changed as discussed

app/controllers/CandidateDetailsController.scala
satishkamatam: Refactored as discussed

app/services/generaldetails/CandidateDetailsService.scala
satishkamatam: As discussed my initial thought was to include fast pass updates in personal repo. But we might need to use the fast pass details else where may be reporting. So Leaving as it.

app/model/FastPassDetails.scala
```diff
@@ -0,0 +1,34 @@
...
+import model.FastPassType.FastPassType
+import play.api.libs.json.Json
+import reactivemongo.bson.Macros
+
+
+case class FastPassDetails(applicable:Boolean,
+                           fastPassType:Option[FastPassType] = None,
+                           internshipTypes:Option[Seq[InternshipType]] = None,
+                           fastPassReceived:Option[Boolean] = None,
+                           certificateNumber:Option[String] = None)
```
henricook: Remarkably minor, but for standardisation reasons alone there should be some spaces after the colons here

app/model/command/GeneralDetailsExchange.scala
```diff
@@ -0,0 +1,38 @@
...
+                                  dateOfBirth: LocalDate,
+                                  outsideUk: Boolean,
+                                  address: Address,
+                                  postCode: Option[PostCode],
+                                  phone: PhoneNumber,
+                                  fastPassDetails: FastPassDetails,
+                                  updateApplicationStatus: Option[Boolean] = None)
+
+object GeneralDetailsExchange {
+  implicit val updateGeneralDetailsFormat = Json.format[GeneralDetailsExchange]
```
henricook: This looks like a missed rename, should probably be generalDetailsExchangeFormat?

henricook:
Back end looks Bon, IMHO :+1:

PR #24, opened 2016-08-31
FSET-706 Groundwork: Add the ability to hard code firstnames, lastnames and preferred names in the test data generator

henricook:


app/services/testdata/RegisteredStatusGenerator.scala
```diff
@@ -24,41 +24,38 @@ import services.testdata.faker.DataFaker._
...
+  override val authProviderClient = AuthProviderClient
 }
 
 trait RegisteredStatusGenerator extends BaseGenerator {
   import scala.concurrent.ExecutionContext.Implicits.global
 
-  val appRepository: GeneralApplicationRepository
-  val pdRepository: PersonalDetailsRepository
-  val qRepository: QuestionnaireRepository
+  val authProviderClient: AuthProviderClient.type
```
kjharkin: Just a question: interesting use of Singleton.type here. What was the intention?

henricook: I think i've used it previously when injecting things that exist as singletons within the app. If I take away the .type things don't work as expected, not an overly scientific explanation! - how would you do this?

kjharkin:
Looks good.

app/services/testdata/RegisteredStatusGenerator.scala
kjharkin: I'm not sure what the issue is if I'm being honest. Does one have to specify Singleton.type when injecting any singleton or only when one wants to limit an interface (trait) to a specific Singleton? Someone I'll need to try.

henricook: I think it's whenever you inject a Singleton, which we don't really do much. I can't find many links to support it though from a cursory look, maybe i'll have a proper google later

PR #25, opened 2016-09-07
FSET-706: Implement candidate search in the admin-frontend

henricook:
- Add first or preferred name search.
 - Return users that are still in REGISTERED from the auth provider.
 - Update test data genny to take post code, dob and email address details

Refactoring:
 - Move logic out of SearchForApplicantController to a service (SearchForApplicantService).
 - When searching by user details and a post code, we now search by post code first so that we then have a list of userIds to query for (rather then previous where we were potentially querying the entire collecton) and can then enrich the result with contact details.


app/connectors/AuthProviderClient.scala
```diff
@@ -47,7 +49,9 @@ trait AuthProviderClient {
 
   val allRoles = List(FaststreamTeamRole, ServiceSupportRole, ServiceAdminRole, SuperAdminRole, TechnicalAdminRole)
 
-  def getRole(roleName: String) = allRoles.find(_.name == roleName).getOrElse(throw new UserRoleDoesNotExist(s"No such role: $roleName"))
+  def getRole(roleName: String) = allRoles.find(_.name == roleName).getOrElse(
```
Lansalo: Wouldn't be better to use a Try as returned type and let the client to deal with the possible exception?
Perhaps out of scope but it's a more functional way to deal with Exception in general.

henricook: So far, in every case this throw is desireable, if we wanted to have different classes using this method dealing with it differently i'd certainly consider using Try - otherwise it's just over engineering IMHO

app/controllers/SearchForApplicantsController.scala
```diff
@@ -41,71 +43,36 @@ trait SearchForApplicantsController extends BaseController {
   val appRepository: GeneralApplicationRepository
   val psRepository: PersonalDetailsRepository
   val cdRepository: ContactDetailsRepository
+  val searchForApplicantService: SearchForApplicantService
 
-  val MAX_RESULTS = 10
+  val MAX_RESULTS = 25
 
   def findById(userId: String, frameworkId: String) = Action.async { implicit request =>
```
Lansalo: I know... out of scope, but if you have a chance, would you mind to add the type returned?

Lansalo: I know... out of scope so you can ignore it. But if you have a chance, would you mind to add the return type to these methods?

app/controllers/testdata/TestDataGeneratorController.scala
```diff
@@ -32,7 +36,11 @@ import uk.gov.hmrc.play.microservice.controller.BaseController
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
-object TestDataGeneratorController extends TestDataGeneratorController
+object TestDataGeneratorController extends TestDataGeneratorController {
+
+  sealed case class InvalidPostCodeFormatException(message: String) extends Exception(message)
```
Lansalo: Why sealed? Maybe final instead?

app/services/search/SearchForApplicantService.scala
```diff
@@ -0,0 +1,124 @@
...
+  val cdRepository: ContactDetailsRepository
+  val authProviderClient: AuthProviderClient
+
+  def findByCriteria(searchCandidate: SearchCandidate)(implicit hc: HeaderCarrier): Future[List[Candidate]] = searchCandidate match {
+    case SearchCandidate(None, None, None, Some(postCode)) => searchByPostCode(postCode)
+
+    case SearchCandidate(firstOrPreferredName, lastName, dateOfBirth, postCode) =>
+      searchByAllNamesOrDobAndFilterPostCode(firstOrPreferredName, lastName, dateOfBirth, postCode)
+
+    case _ => Future.successful(Nil)
```
Lansalo: In this case, more the fact that we didn't find any result, we would like to flag the fact that we couldn't perform any search for the given inputs... so wouldn't a Future.failed(<SomeException>) be more appropriate?

app/services/search/SearchForApplicantService.scala
```diff
@@ -0,0 +1,124 @@
...
+        appRepository.findCandidateByUserId(cd.userId).map(_.map { candidate =>
+          candidate.copy(address = Some(cd.address), postCode = Some(cd.postCode))
+        }).recover {
+          case e: ContactDetailsNotFound => None
+        }
+      })
+    }.map(_.flatten)
+  }
+
+  private def searchByAllNamesOrDobAndFilterPostCode(firstOrPreferredName: Option[String],
```
Lansalo: Return type

app/services/search/SearchForApplicantService.scala
```diff
@@ -0,0 +1,124 @@
...
+  def findByCriteria(searchCandidate: SearchCandidate)(implicit hc: HeaderCarrier): Future[List[Candidate]] = searchCandidate match {
+    case SearchCandidate(None, None, None, Some(postCode)) => searchByPostCode(postCode)
+
+    case SearchCandidate(firstOrPreferredName, lastName, dateOfBirth, postCode) =>
+      searchByAllNamesOrDobAndFilterPostCode(firstOrPreferredName, lastName, dateOfBirth, postCode)
+
+    case _ => Future.successful(Nil)
+  }
+
+  private def searchByPostCode(postCode: String) = {
```
Lansalo: return type

app/services/testdata/GeneratorConfig.scala
```diff
@@ -17,9 +17,10 @@
 package services.testdata
 
 import model.EvaluationResults.Result
+import org.joda.time.LocalDate
```
Lansalo: Maybe I'm wrong of just dream of it, but I read somewhere to stop using Joda. Maybe Olly knows a bit more?

app/services/search/SearchForApplicantService.scala
henricook: Do we know how play deals with a failed future? I agree this should be a bad input/error condition because presumably we've received a Search Candidate object that's somehow malformed. I'm not even entirely sure that this default case is reachable

henricook: I think it is unreachable, i've removed the default case and pushed

app/services/testdata/GeneratorConfig.scala
henricook: Yep, we want to stop using it but we can't while we're still on Play 2.3 - at least not easily. There's an internal HMRC spike happening to upgrade our libraries etc to play 2.5 - so eventually it will be possible.

app/services/search/SearchForApplicantService.scala
henricook: This is done

PR #26, opened 2016-09-08
FSET-759: View a candidate

satishkamatam:


app/model/command/GeneralDetails.scala
```diff
@@ -21,19 +21,19 @@ import model.Commands.{ PhoneNumber, PostCode }
...
+                          postCode: Option[PostCode],
+                          country: Option[String],
+                          phone: PhoneNumber,
+                          fastPassDetails: FastPassDetails,
+                          updateApplicationStatus: Option[Boolean] = None)
 
-object GeneralDetailsExchange {
-  implicit val updateGeneralDetailsFormat = Json.format[GeneralDetailsExchange]
+object GeneralDetails {
+  implicit val updateGeneralDetailsFormat = Json.format[GeneralDetails]
```
mgilgar: I would call this generalDetailsFormat.

app/controllers/ApplicationController.scala
```diff
@@ -60,8 +59,8 @@ trait ApplicationController extends BaseController {
   }
 
   def applicationStatus(applicationId: String) = Action.async { implicit request =>
-    appRepository.findProgress(applicationId).map { result =>
-      Ok(Json.toJson(ApplicationStatusOrder.getStatus(result)))
+    appRepository.findStatus(applicationId).map { result =>
```
mgilgar: I would call ApplicationController.applicationStatus() the same as ApplicationRepository.findStatus. My suggestion would be aplicationStatusDetails as this is what we are dealing with, and status or applicationStatus refers more to the status string.

app/controllers/ApplicationController.scala
```diff
@@ -60,8 +59,8 @@ trait ApplicationController extends BaseController {
...
+      Ok(Json.toJson(result))
     }.recover {
       case e: ApplicationNotFound => NotFound(s"cannot find application for user with id: ${e.id}")
```
mgilgar: Not sure if the exception message is appropriate because what we want is the application status, so we fail and we say application not found, that might be confusing. I would say cannot retrieve applications status details for application with id=$applicationId. Instead of the userId, I think applicationId is more relevant in this case.

app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -319,47 +334,37 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService)(implic
...
-      "withdraw" -> reason,
-      "applicationStatus" -> "WITHDRAWN",
-      "progress-status.withdrawn" -> true
-    ))
-
-    collection.update(query, applicationStatusBSON, upsert = false) map {
-      case _ => ()
-    }
+    val applicationBSON = BSONDocument("$set" -> BSONDocument(
+        "withdraw" -> reason
```
mgilgar: It was there before but I think it should be something like "withdrawReason" instead of "withdraw"

app/controllers/ApplicationController.scala
satishkamatam: done

app/controllers/ApplicationController.scala
satishkamatam: changed to findApplicationStatusDetails

app/repositories/application/GeneralApplicationRepository.scala
satishkamatam: Agreed. But this is a low level model change, I would rather defer this to withdraw story.

mgilgar: That's fine.

PR #30, opened 2016-09-12
New Progress statuses ahead of new cubiks integration

henricook:


app/model/ApplicationStatuses.scala
```diff
@@ -23,10 +23,6 @@ object ApplicationStatuses {
   val Withdrawn = "WITHDRAWN"
```
lukastymo: We can add ```@deprecated``` on the top of this file to show clearly that we shouldn't add new statuses here.

app/model/ProgressStatuses.scala
```diff
@@ -47,4 +47,40 @@ object ProgressStatuses {
...
+
+  case object PHASE1_TESTS_STARTED extends ProgressStatus(ApplicationStatus.PHASE1_TESTS)
+
+  case object PHASE1_TESTS_COMPLETED extends ProgressStatus(ApplicationStatus.PHASE1_TESTS)
+
+  case object PHASE1_TESTS_EXPIRED extends ProgressStatus(ApplicationStatus.PHASE1_TESTS)
+
+  case object PHASE1_TESTS_RESULTS_RECEIVED extends ProgressStatus(ApplicationStatus.PHASE1_TESTS)
+
+  val nameToProgressStatus = List(
```
lukastymo: - add private
- There is some problem with the type (intellij show an error in the line where we use this map). To solve this you can try to add explicit type: ```Map[String, ProgressStatus]```

app/model/ProgressStatuses.scala
```diff
@@ -47,4 +47,40 @@ object ProgressStatuses {
   val AssessmentCentreFailedProgress = ApplicationStatuses.AssessmentCentreFailed.toLowerCase()
   val AssessmentCentrePassedNotifiedProgress = ApplicationStatuses.AssessmentCentrePassedNotified.toLowerCase()
   val AssessmentCentreFailedNotifiedProgress = ApplicationStatuses.AssessmentCentreFailedNotified.toLowerCase()
+
+  sealed abstract class ProgressStatus(applicationStatus: ApplicationStatus)
+
+  object ProgressStatus {
+    implicit val progressStatusFormat = new Format[ProgressStatus] {
+      def reads(json: JsValue) = JsSuccess(nameToProgressStatus(json.as[String]))
+      def writes(myEnum: ProgressStatus) = JsString(myEnum.toString)
```
lukastymo: it's not enum anymore: ```myEnum```

app/model/ProgressStatuses.scala
henricook: Done

app/model/ProgressStatuses.scala
henricook: Done

app/model/ApplicationStatuses.scala
henricook: Done

app/model/ProgressStatuses.scala
```diff
@@ -47,4 +47,40 @@ object ProgressStatuses {
...
+    }
+
+    implicit object BSONEnumHandler extends BSONHandler[BSONString, ProgressStatus] {
+      def read(doc: BSONString) = nameToProgressStatus(doc.value)
+      def write(stats: ProgressStatus) = BSON.write(stats.toString)
+    }
+  }
+
+  case object PHASE1_TESTS_INVITED extends ProgressStatus(ApplicationStatus.PHASE1_TESTS)
```
lukastymo: We need to make sure that everybody agree with the new capital case convention for progress statuses + we need refactor the old ones

henricook: @ollyw How do you feel about new progress statuses in capitals? If nothing else it helps us easily spot if we're using a "new" progress status or an "old" one

PR #31, opened 2016-09-12
FSET-766 Invite candidate to online tests

benjstephenson:


conf/application.conf
```diff
@@ -92,8 +92,8 @@ microservice {
     cubiks-gateway {
       url = "http://localhost:9288"
       scheduleIds {
-        standard = 15547
-        gis = 15593
+        standard = [ 15547, 12345 ]
```
henricook: Because etray is also going to be a Cubiks test there should probably something that marks these as 'phase1' here - phase1 { standard []  gis [] } or phase1-standard perhaps?

conf/application.conf
```diff
@@ -135,7 +135,7 @@ microservice {
 scheduling {
   online-testing {
     send-invitation-job {
-      enabled = false
+      enabled = true
```
henricook: We keep all the schedulers in dev off by default - so please amend this to false before putting up for review

app/services/testdata/OnlineTestInvitedStatusGenerator.scala
```diff
@@ -43,7 +43,8 @@ trait OnlineTestInvitedStatusGenerator extends ConstructiveGenerator {
       onlineTestUrl = generatorConfig.cubiksUrl,
       invitationDate = DateTime.now(),
       expirationDate = DateTime.now().plusDays(7),
-      participantScheduleId = 149245
+      participantScheduleId = 149245,
```
mgilgar: Last time, in fasttrack, we added "participantScheduleId" because it was returned by cubiks but finally we did not use it but we keep it because of time constraints. It might be worth to revisit this and remove it if it is not necessary.

ollyw: @mgilgar our plan going forward is to try and keep all data that Cubiks, and other integration partners return, to help with diagnostics.

test/services/onlinetesting/OnlineTestServiceSpec.scala
```diff
@@ -187,11 +182,70 @@ class OnlineTestServiceSpec extends PlaySpec with BeforeAndAfterEach with Mockit
   }
 
   "register and invite application" should {
+
+    "issue one email for invites to SJQ for GIS candidates" in new OnlineTest {
+
```
mgilgar: Complex tests may be a hint of having complicate code that needs to be refactored. I know sometimes it is impossible to make it smaller but it is worth to revisit this and see if we can improve it in the original service and then, these tests

test/services/onlinetesting/OnlineTestServiceSpec.scala
```diff
@@ -218,88 +272,87 @@ class OnlineTestServiceSpec extends PlaySpec with BeforeAndAfterEach with Mockit
...
-        verify(auditServiceMock).logEventNoRequest("UserInvitedToOnlineTest", auditDetails)
-        verify(auditServiceMock, times(2)).logEventNoRequest(any[String], any[Map[String, String]])
+        verify(auditServiceMock, times(2)).logEventNoRequest("UserRegisteredForOnlineTest", auditDetails)
+        verify(auditServiceMock, times(2)).logEventNoRequest("UserInvitedToOnlineTest", auditDetails)
+        verify(auditServiceMock, times(4)).logEventNoRequest(any[String], any[Map[String, String]])
       }
     "fail, audit 'UserRegisteredForOnlineTest' and audit 'UserInvitedToOnlineTest'" +
       " if there is an exception sending the invitation email" in new OnlineTest {
-        when(cubiksGatewayClientMock.registerApplicant(eqTo(registerApplicant))(any[HeaderCarrier]))
+        when(cubiksGatewayClientMock.registerApplicant(any[RegisterApplicant])(any[HeaderCarrier]))
```
mgilgar: It should work with "registerApplicant".

test/services/onlinetesting/OnlineTestServiceSpec.scala
```diff
@@ -218,88 +272,87 @@ class OnlineTestServiceSpec extends PlaySpec with BeforeAndAfterEach with Mockit
...
           .thenReturn(Future.successful(registration))
-        when(cubiksGatewayClientMock.inviteApplicant(eqTo(inviteApplicant))(any[HeaderCarrier]))
+        when(cubiksGatewayClientMock.inviteApplicant(any[InviteApplicant])(any[HeaderCarrier]))
```
mgilgar: It should work with "inviteApplicant".

test/services/onlinetesting/OnlineTestServiceSpec.scala
```diff
@@ -218,88 +272,87 @@ class OnlineTestServiceSpec extends PlaySpec with BeforeAndAfterEach with Mockit
...
+        verify(auditServiceMock, times(4)).logEventNoRequest(any[String], any[Map[String, String]])
       }
     "fail, audit 'UserRegisteredForOnlineTest', audit 'UserInvitedToOnlineTest'" +
-      ", send invitation email to user" +
+      ", not send invitation email to user" +
       "if there is an exception storing the status and the online profile data to database" in new OnlineTest {
         when(cubiksGatewayClientMock.registerApplicant(eqTo(registerApplicant))(any[HeaderCarrier]))
           .thenReturn(Future.successful(registration))
-        when(cubiksGatewayClientMock.inviteApplicant(eqTo(inviteApplicant))(any[HeaderCarrier]))
+        when(cubiksGatewayClientMock.inviteApplicant(any[InviteApplicant])(any[HeaderCarrier]))
```
mgilgar: It should work with "inviteApplicant".

test/services/onlinetesting/OnlineTestServiceSpec.scala
```diff
@@ -218,88 +272,87 @@ class OnlineTestServiceSpec extends PlaySpec with BeforeAndAfterEach with Mockit
...
+        verify(auditServiceMock, times(2)).logEventNoRequest("UserRegisteredForOnlineTest", auditDetails)
+        verify(auditServiceMock, times(2)).logEventNoRequest("UserInvitedToOnlineTest", auditDetails)
+        verify(auditServiceMock, times(0)).logEventNoRequest("OnlineTestInvitationEmailSent", auditDetailsWithEmail)
+        verify(auditServiceMock, times(4)).logEventNoRequest(any[String], any[Map[String, String]])
       }
     "audit 'OnlineTestInvitationProcessComplete' on success" in new OnlineTest {
       when(cubiksGatewayClientMock.registerApplicant(eqTo(registerApplicant))(any[HeaderCarrier]))
         .thenReturn(Future.successful(registration))
-      when(cubiksGatewayClientMock.inviteApplicant(eqTo(inviteApplicant))(any[HeaderCarrier]))
+      when(cubiksGatewayClientMock.inviteApplicant(any[InviteApplicant])(any[HeaderCarrier]))
```
mgilgar: Generally speaking, it should work with inviteApplication or registerApplicant. It used to work at least.

benjstephenson:
@mgilgar I didn't creating new instances of the InviteApplicant and RegisterApplicant case classes for the eq matchers as I think all that is really doing is testing the mock of the cubiks gateway.  The behaviour of that class will be tested in it's own test spec.

henricook:
Restarting this in travis

app/controllers/OnlineTestController.scala
```diff
@@ -45,34 +46,34 @@ case class OnlineTestExtension(extraDays: Int)
...
   import Commands.Implicits._
 
   def getOnlineTest(userId: String) = Action.async { implicit request =>
-    onlineTestingService.getOnlineTest(userId).map { onlineTest =>
-      Ok(Json.toJson(onlineTest))
+
+    onlineTestingService.getPhase1TestProfile(userId).map {
+      case Some(phase1TestProfile) => Ok(Json.toJson(phase1TestProfile))
+      case None => NotFound
     } recover {
```
henricook: We're effectively hiding an exception here from the frontends, we should probably log it

henricook: This method needs a return type according to our new style guidance (Intellij will add it for you)

app/controllers/OnlineTestController.scala
```diff
@@ -85,40 +86,40 @@ trait OnlineTestController extends BaseController {
    * @return
    */
   def completeTests(token: String) = Action.async { implicit request =>
```
henricook: In theory these also need return types, but since they're not the focus of your PR they'll be optional bonus points only

app/controllers/OnlineTestController.scala
```diff
@@ -45,34 +46,34 @@ case class OnlineTestExtension(extraDays: Int)
...
 
+  val applicationRepository: GeneralApplicationRepository
   val onlineRepository: OnlineTestRepository
   val onlineTestingService: OnlineTestService
   val onlineTestExtensionService: OnlineTestExtensionService
-  val onlineTestPDFReportRepo: OnlineTestPDFReportRepository
 
   import Commands.Implicits._
 
   def getOnlineTest(userId: String) = Action.async { implicit request =>
```
henricook: This method needs a return type according to our new style guidance (Intellij will add it for you)

app/model/ApplicationStatus.scala
```diff
@@ -24,6 +24,8 @@ object ApplicationStatus extends Enumeration {
   val WITHDRAWN, CREATED, IN_PROGRESS, SUBMITTED = Value
   val PHASE1_TESTS = Value
 
+  implicit def toString(applicationStatus: ApplicationStatus): String = applicationStatus.toString
```
henricook: I don't think you need this, enum's have a default toString - which you're using :-)

app/repositories/CommonBSONDocuments.scala
```diff
@@ -16,12 +16,22 @@
...
 package repositories
 
+import model.{ApplicationStatus, ProgressStatuses}
 import org.joda.time.LocalDate
 import reactivemongo.bson.BSONDocument
 
 
 trait CommonBSONDocuments {
 
+  def applicationStatusBSON( progressStatus: ProgressStatuses.ProgressStatus): BSONDocument = {
```
henricook: I mean it's world changing feedback, but there's a leading space here

app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -24,16 +24,17 @@ import model.ApplicationStatuses._
...
 import model.PersistedObjects.ApplicationForNotification
 import model._
 import model.command._
 import org.joda.time.format.DateTimeFormat
-import org.joda.time.{ DateTime, LocalDate }
+import org.joda.time.{DateTime, LocalDate}
 import play.api.Logger
-import play.api.libs.json.{ Format, JsNumber, JsObject }
-import reactivemongo.api.{ DB, QueryOpts, ReadPreference }
-import reactivemongo.bson.{ BSONDocument, _ }
```
henricook: I believe we decided on spaces before and after curly braces in imports, it's not a default intellij option (but it is the scalariform default) - you'll need to enable this option if i've remembered it the right way round:

selection_133

app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -998,6 +1007,74 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService)(implic
     collection.update(query, passMarkEvaluation, upsert = false) map { _ => }
   }
 
+  // TODO: get rid of this, it feels wrong and we should also be using the App Status case objects.
+  private def applicationStatus(status: String): BSONDocument = {
```
henricook: Can i have some backstory, where did this appear from? It uses all the old defunct app statuses, so should be good to comment or remove

app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -998,6 +1007,74 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService)(implic
...
+
+  private def initialisePhase1TestStatus(applicationId: String): Future[Unit] = {
+    import model.ProgressStatuses._
+
+    val query = BSONDocument("applicationId" -> applicationId)
+
+    val applicationStatusBSON = BSONDocument("$unset" -> BSONDocument(
+      s"progress-status.$PHASE1_TESTS_COMPLETED" -> "",
+      s"progress-status.$PHASE1_TESTS_EXPIRED" -> "",
+      s"progress-status.$AwaitingOnlineTestReevaluationProgress" -> "",
```
henricook: I'd bin these 4 states, they're not being used yet and will all be refactored

henricook: And passmarkEvaluation

app/repositories/application/OnlineTestRepository.scala
```diff
@@ -151,49 +100,21 @@ class OnlineTestMongoRepository(dateTime: DateTimeFactory)(implicit mongo: () =>
...
-      "applicationStatus" -> "ONLINE_TEST_INVITED",
-      "online-tests.cubiksUserId" -> onlineTestProfile.cubiksUserId,
-      "online-tests.token" -> onlineTestProfile.token,
-      "online-tests.onlineTestUrl" -> onlineTestProfile.onlineTestUrl,
-      "online-tests.invitationDate" -> onlineTestProfile.invitationDate,
-      "online-tests.expirationDate" -> onlineTestProfile.expirationDate,
-      "online-tests.participantScheduleId" -> onlineTestProfile.participantScheduleId
-    ))
+    val doc = BSONDocument("applicationId" -> applicationId,
+      "testGroups" -> BSONDocument("PHASE1" -> phase1TestProfile)
```
henricook: Just a comment: we should probably put "PHASE1" into a constant at some point

app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -173,27 +173,31 @@ trait OnlineTestService {
...
   }
 
-  private def markAsCompleted(application: OnlineTestApplication)(onlineTestProfile: OnlineTestProfile): Future[Unit] =
-    otRepository.storeOnlineTestProfileAndUpdateStatusToInvite(application.applicationId, onlineTestProfile).map { _ =>
-      audit("OnlineTestInvitationProcessComplete", application.userId)
-    }
+  private def markAsCompleted(application: OnlineTestApplication)
+    (onlineTestProfile: Phase1TestProfile): Future[Unit] = for {
+    _ <- otRepository.insertPhase1TestProfile(application.applicationId, onlineTestProfile)
+    _ <- appRepository.updateProgressStatus(application.applicationId, ProgressStatuses.PHASE1_TESTS_INVITED)
```
henricook: This method looks like it should be marked as COMPLETE rather than INVITED?

app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -173,27 +173,31 @@ trait OnlineTestService {
...
+  } yield {
+      audit(s"ApplicationStatus set to ${ApplicationStatus.PHASE1_TESTS} - ProgressStatus set to" +
```
henricook: There might be a reason we only use these one word event audits like 'OnlineTestInvitationEmailSent' - is there a reason you know of @ollyw for the one word audits rather than string messages like this?

app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -253,12 +259,22 @@ trait OnlineTestService {
   }
 
   private[services] def buildInviteApplication(application: OnlineTestApplication, token: String, userId: Int, scheduleId: Int) = {
-    val onlineTestCompletedUrl = gatewayConfig.candidateAppUrl + "/fset-fast-stream/online-tests/complete/" + token
+    val scheduleCompletionUrl = gatewayConfig.candidateAppUrl + "/fset-fast-stream/online-tests/complete/" + token
```
henricook: We should probably change this route, do you think? phase1-tests or something?

test/services/onlinetesting/OnlineTestServiceSpec.scala
```diff
@@ -136,62 +150,107 @@ class OnlineTestServiceSpec extends PlaySpec with BeforeAndAfterEach with Mockit
...
     }
   }
 
   "register and invite application" should {
+
+    "issue one email for invites to SJQ for GIS candidates" in new OnlineTest {
+
+      when(cubiksGatewayClientMock.registerApplicant(eqTo(registerApplicant))(any[HeaderCarrier]))
+        .thenReturn(Future.successful(registration))
+      when(cubiksGatewayClientMock.inviteApplicant(any[InviteApplicant])(any[HeaderCarrier]))
```
henricook: I'd probably move this mock setup into a testfixture trait at the bottom of the file, leaves the test clean and readable

test/services/onlinetesting/OnlineTestServiceSpec.scala
```diff
@@ -136,62 +150,107 @@ class OnlineTestServiceSpec extends PlaySpec with BeforeAndAfterEach with Mockit
...
+      verify(auditServiceMock, times(1)).logEventNoRequest("OnlineTestInvitationEmailSent", auditDetailsWithEmail)
+      verify(auditServiceMock, times(1)).logEventNoRequest("OnlineTestInvitationProcessComplete", auditDetails)
+      verify(auditServiceMock, times(1)).logEventNoRequest(progressStatusAuditMessage, auditDetails)
+      verify(auditServiceMock, times(5)).logEventNoRequest(any[String], any[Map[String, String]])
+    }
+
+    "issue one email for invites to SJQ and BQ tests for non GIS candidates" in new OnlineTest {
+
+      when(cubiksGatewayClientMock.registerApplicant(eqTo(registerApplicant))(any[HeaderCarrier]))
+        .thenReturn(Future.successful(registration))
```
henricook: Ditto on the mock setup

app/repositories/application/GeneralApplicationRepository.scala
benjstephenson: Removed as it's not used - using the applciationStatusBSON function to consistently set the progress, application status and progress status date fields instead.

app/controllers/OnlineTestController.scala
benjstephenson: Removed the recover so the exception is propagated to the caller.

app/controllers/OnlineTestController.scala
benjstephenson: I thought the result of our convo on it with @ollyw the other week was that we didn't really need them on controller methods?

app/model/ApplicationStatus.scala
benjstephenson: They do, but it's not implicit and it was mostly so I didn't have to keep calling ```toString``` on variables

app/repositories/CommonBSONDocuments.scala
benjstephenson: :scream:

app/services/onlinetesting/OnlineTestService.scala
benjstephenson: Renamed

app/services/onlinetesting/OnlineTestService.scala
benjstephenson: I don't think so.  This endpoint is just concerned with completing a test using a token as a key.  If we had endpoints for each test the implementation for them would be mostly the same.

PR #35, opened 2016-09-16
Move online test to application to allow queries with Application Status

lukastymo:


app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -1016,22 +1017,14 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService)(implic
...
+    collection.update(query, BSONDocument("$set" ->
+      applicationStatusBSON(progressStatus))
+    ) map { _ => }
   }
 
-
-  private def initialisePhase1TestStatus(applicationId: String): Future[Unit] = {
-    import model.ProgressStatuses._
-
+  override def insertPhase1TestProfile(applicationId: String, phase1TestProfile: Phase1TestProfile) = {
```
ollyw: @lukastymo Let's catch up about this. I see that insertPhase1TestProfile is in both OnlineTestRepository and GeneralApplicationRepository now. The duplication needs to be resolved. Also previously we have created multiple repositories that use the application collection (in the repositories/application) folder. Therefore I was expecting that the test info access methods would be in OnlineTestRepository in repositories/application. We originally added too many methods to a single repository for the application collection and it wasn't pleasant. I am keen not to do this again, unless there is something I am missing that requires it (which there may be!)

PR #37, opened 2016-09-16
FSET-641: Add reset test service method

lukastymo:


app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -77,10 +77,8 @@ trait OnlineTestService {
 
   def getPhase1TestProfile(userId: String): Future[Option[Phase1TestProfile]] = {
     appRepository.findCandidateByUserId(userId).flatMap {
-      case Some(candidate) if candidate.applicationId.isDefined =>
-        otRepository.getPhase1TestProfile(candidate.applicationId.get)
+      case Some(candidate) => otRepository.getPhase1TestProfile(candidate.applicationId.getOrElse(""))
```
ollyw: Under what circumstance would you want to get a getPhase1TestProfile if the candidate didn't actually have an application? If there isn't any circumstance that make sense, it seems to me that a plain old .get is better here, or the guard.

app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -183,20 +181,19 @@ trait OnlineTestService {
...
   private def markAsInvited(application: OnlineTestApplication)
-    (onlineTestProfile: Phase1TestProfile): Future[Unit] = for {
-    _ <- otRepository.insertPhase1TestProfile(application.applicationId, onlineTestProfile)
-    _ <- appRepository.updateProgressStatus(application.applicationId, ProgressStatuses.PHASE1_TESTS_INVITED)
+                           (onlineTestProfile: Phase1TestProfile): Future[Unit] = for {
+    _ <- appRepository.insertPhase1TestProfile(application.applicationId, onlineTestProfile)
   } yield {
-      audit(s"ApplicationStatus set to ${ApplicationStatus.PHASE1_TESTS} - ProgressStatus set to" +
-        s" ${ProgressStatuses.PHASE1_TESTS_INVITED}", application.userId)
+    audit(s"ApplicationStatus set to ${ApplicationStatus.PHASE1_TESTS} - ProgressStatus set to" +
```
ollyw: the first parameter of audit is the event id, not a free text description. It should be something like OnlineTestInvited or whatever is relevant. If additional meta data is needed, it can be passed into logEventNoRequest()

app/repositories/application/OnlineTestRepository.scala
```diff
@@ -66,17 +64,18 @@ trait OnlineTestRepository {
...
 
   override def getPhase1TestProfile(applicationId: String): Future[Option[Phase1TestProfile]] = {
-
     val query = BSONDocument("applicationId" -> applicationId)
     val projection = BSONDocument("testGroups.PHASE1" -> 1, "_id" -> 0)
 
-    collection.find(query, projection).one[BSONDocument].map {
-      case Some(doc) => doc.getAs[Phase1TestProfile]("testGroups.PHASE1").orElse(None)
+    collection.find(query, projection).one[BSONDocument] map {
+      case Some(doc) if doc.getAs[BSONDocument]("testGroups").isDefined => Some(
```
ollyw: maybe a map is a bit more eloquent then isDefined followed by a get

app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -183,20 +186,38 @@ trait OnlineTestService {
...
-    _ <- otRepository.insertPhase1TestProfile(application.applicationId, onlineTestProfile)
-    _ <- appRepository.updateProgressStatus(application.applicationId, ProgressStatuses.PHASE1_TESTS_INVITED)
+                           (newOnlineTestProfile: Phase1TestProfile): Future[Unit] = for {
+    currentOnlineTestProfile <- otRepository.getPhase1TestProfile(application.applicationId)
+    updatedOnlineTestProfile = merge(currentOnlineTestProfile, newOnlineTestProfile)
+    _ <- appRepository.insertPhase1TestProfile(application.applicationId, updatedOnlineTestProfile)
   } yield {
-      audit(s"ApplicationStatus set to ${ApplicationStatus.PHASE1_TESTS} - ProgressStatus set to" +
-        s" ${ProgressStatuses.PHASE1_TESTS_INVITED}", application.userId)
+    audit(s"ApplicationStatus set to ${ApplicationStatus.PHASE1_TESTS} - ProgressStatus set to" +
```
ollyw: Again, audit requires an event Id, not free text

app/services/onlinetesting/OnlineTestService.scala
lukastymo: I'll add the guard

app/services/onlinetesting/OnlineTestService.scala
lukastymo: The code is from the other branch (merge) - I'll update this here, but we need to be careful when merging later to not override the fix

PR #39, opened 2016-09-18
Fixup some futures, method renaming and more explicit assertions

ollyw:
Most of this code has migrated to other PRs, so this is a very small bunch of cleanups

### DIFF
--- a/app/services/onlinetesting/OnlineTestService.scala
+++ b/app/services/onlinetesting/OnlineTestService.scala
@@ -158,8 +158,6 @@ trait OnlineTestService {
         testUrl = invitation.authenticateUrl
       )
     }
-
-
   }
 
   def retrieveTestResult(application: OnlineTestApplicationWithCubiksUser, waitSecs: Option[Int]): Future[Unit] = {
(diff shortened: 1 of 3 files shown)


app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -85,10 +85,12 @@ trait OnlineTestService {
...
 
   def registerAndInviteForTestGroup(application: OnlineTestApplication): Future[Unit] = {
-    val (invitationDate, expirationDate) = onlineTestDates
-    val registerAndInviteProcess = Future.sequence(getScheduleIdForApplication(application).map { scheduleId =>
-      registerAndInviteApplicant(application, scheduleId, invitationDate, expirationDate)
-    }).map { phase1Tests =>
+    val (invitationDate, expirationDate) = calcOnlineTestDates()
+
+    val registerAndInviteProcess = Future.sequence(
+      getScheduleIdsForApplication(application).map { scheduleId =>
```
Lansalo: Any particular reason why we use Future.sequence(...) rather than simply Future(...)?

ollyw: Because the getScheduleIdsForApplication returns multiple scheduleIds, and therefore we create multiple futures, and we need to wait for all the futures, so using Future.sequence turns them into a single future to map further operations

PR #42, opened 2016-09-19
FSET 642 - WIP. Important base methods for working with online tests needed by others.

henricook:
WIP - tests pass, still a lot to do but this should release other people's reliance on my branch.

- Mainly: Add ability to generate users 'in test' with start and expiry times with the test data generator
- Add controller for 'phase1 test groups' and stubbed extend methods
- Repository methods to extend test group expiry (untested, tests coming)
- Progress statuses for dealing with online tests
- Stubbed service method for extending (impl coming)


mgilgar:
It looks good but I have not reviewed throughly what it is not related to the test data generator.

henricook:
I'd say if tests pass (which they do from the travis run) lets merge it - as long as there's nothing in there that's not clearly outrageous, we'll review the entire cubiks piece after the fact as well I would guess

PR #43, opened 2016-09-19
Fset 640: Expire a candidate's online exercises

Lansalo:


Lansalo:
I still would like to do some more work around the test coverage but it's probably low priority so I'll focus on the email template first.

henricook:
@Lansalo Looks like there's some conflicts here, could we also have a more descriptive PR title please? E.g. "FSET 640 - Making a button to summon unicorns" Vs. just "FSET 640"

app/repositories/application/OnlineTestRepository.scala
```diff
@@ -46,6 +46,8 @@ trait OnlineTestRepository {
 
   def insertOrUpdatePhase1TestGroup(applicationId: String, phase1TestProfile: Phase1TestProfile): Future[Unit]
 
+  def nextApplicationPendingExpiry: Future[Option[ExpiringOnlineTest]]
```
mgilgar: Why don't we call the method "nextExpiringApplication"?

app/repositories/application/OnlineTestRepository.scala
```diff
@@ -46,6 +46,8 @@ trait OnlineTestRepository {
```
mgilgar: This is a general comment, will be have a separate repository for the different online tests: video interview, etc?

If yes, maybe we should renamed all things related to phase1 to something different maybe OnlineExercise or Phase1OnlineTest?

Anyway I would recommend to split everything in 3.

app/services/onlinetesting/OnlineTestExpiryService.scala
```diff
@@ -42,16 +43,13 @@ class OnlineTestExpiryServiceImpl(
   newHeaderCarrier: => HeaderCarrier
 )(implicit executor: ExecutionContext) extends OnlineTestExpiryService {
 
-  private final val ExpiredStatus = "ONLINE_TEST_EXPIRED"
   private implicit def headerCarrier = newHeaderCarrier
 
   override def processNextExpiredTest(): Future[Unit] = {
```
mgilgar: I think if the name of the method is "processNextExpiredTest", then this line:

otRepository.nextApplicationPendingExpiry

should be:
otRepository.nextApplicationExpiredTest

nextApplicationPendingExpiry is supposed to return an application (by the name), if it is an online tests what is returning, it should say: nextExpiredOnlineTest.

app/services/onlinetesting/OnlineTestExpiryService.scala
```diff
@@ -42,16 +43,13 @@ class OnlineTestExpiryServiceImpl(
...
-    // TODO FAST STREAM FIX ME
-    Future.successful(Unit)
-    //otRepository.nextApplicationPendingExpiry.flatMap {
-    //  case Some(expiringTest) => processExpiredTest(expiringTest)
-    //  case None => Future.successful(())
-    //}
+    otRepository.nextApplicationPendingExpiry.flatMap {
+      case Some(expiringTest) => processExpiredTest(expiringTest)
```
mgilgar: if we get an expiringTest, as I wrote before, it is weird that the method is called nextApplication....

app/services/onlinetesting/OnlineTestExpiryService.scala
```diff
@@ -67,7 +65,7 @@ class OnlineTestExpiryServiceImpl(
     }
 
   override def commitExpiredStatus(expiringTest: ExpiringOnlineTest): Future[Unit] =
-    applicationRepository.updateStatus(expiringTest.userId, ExpiredStatus).map { _ =>
```
mgilgar: In this case I think it should say commietExpiredProgressStatus

app/repositories/application/OnlineTestRepository.scala
Lansalo: Don't think will have time given the deadline. Do you want to track this somewhere?

app/repositories/application/OnlineTestRepository.scala
Lansalo: No particular reason apart from the fact that that was the name in FastTrack.
Happy to change it though.

app/services/onlinetesting/OnlineTestExpiryService.scala
Lansalo: You mean "nextExpiringApplication" like in the comment above right?
I assume so as it's the same method and also at this stage the application is not really expired yet.

app/services/onlinetesting/OnlineTestExpiryService.scala
Lansalo: Done

app/services/onlinetesting/OnlineTestExpiryService.scala
Lansalo: Done

PR #46, opened 2016-09-20
FSET-804 Use correct email template for online tests invitation.

benjstephenson:


henricook:
Now needs to be in HMRC/fset-faststream-email

PR #48, opened 2016-09-20
Add service methods for online test: started/completed/report ready

lukastymo:


app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -323,6 +325,52 @@ trait OnlineTestService {
...
+        case t if t.cubiksUserId == cubiksUserId => update(t)
+        case t => t
+      }
+      val updatedTestGroup = testGroup.copy(tests = updatedTestsWithTodaysDate)
+      (appId, updatedTestGroup)
+    }
+
+    for {
+      p <- otRepository.getPhase1TestProfileByCubiksId(cubiksUserId)
+      (appId, updatedTestGroup) = createUpdateTestGroup(p)
```
ollyw: There is a potential race condition here. Did we look at the possibility of using findAndModify rather than pulling the record back to the service?

PR #55, opened 2016-09-21
Schedule completion URLs for Cubiks

benjstephenson:
Provide Cubiks with different schedule completion URLs depending on GIS status and schedule Id so we can render different status pages.

### DIFF
--- a/app/controllers/OnlineTestController.scala
+++ b/app/controllers/OnlineTestController.scala
@@ -65,7 +65,7 @@ trait OnlineTestController extends BaseController {
   def getOnlineTest(userId: String) = Action.async { implicit request =>
     onlineTestingService.getPhase1TestProfile(userId).map {
       case Some(phase1TestProfileWithNames) => Ok(Json.toJson(phase1TestProfileWithNames))
-      case None => Logger.error(s"No phase 1 test found for userID [$userId]")
+      case None => Logger.info(s"No phase 1 test found for userID [$userId]")
         NotFound
     }
   }
--- a/app/services/onlinetesting/OnlineTestService.scala
+++ b/app/services/onlinetesting/OnlineTestService.scala
@@ -103,9 +103,9 @@ trait OnlineTestService {
     }
   }
 
-  private def sjq = gatewayConfig.onlineTestConfig.scheduleIds("sjq")
+  private def sjq = gatewayConfig.phase1TestsConfig.scheduleIds("sjq")
 
-  private def bq = gatewayConfig.onlineTestConfig.scheduleIds("bq")
+  private def bq = gatewayConfig.phase1TestsConfig.scheduleIds("bq")
 
   def registerAndInviteForTestGroup(application: OnlineTestApplication): Future[Unit] = {
     registerAndInviteForTestGroup(application, getScheduleNamesForApplication(application))
@@ -263,7 +263,7 @@ trait OnlineTestService {
 
   private def calcOnlineTestDates: (DateTime, DateTime) = {
     val invitationDate = dateTimeFactory.nowLocalTimeZone
-    val expirationDate = invitationDate.plusDays(gatewayConfig.onlineTestConfig.expiryTimeInDays)
+    val expirationDate = invitationDate.plusDays(gatewayConfig.phase1TestsConfig.expiryTimeInDays)
     (invitationDate, expirationDate)
   }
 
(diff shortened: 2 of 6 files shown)


app/controllers/OnlineTestController.scala
```diff
@@ -65,7 +65,7 @@ trait OnlineTestController extends BaseController {
   def getOnlineTest(userId: String) = Action.async { implicit request =>
```
mgilgar: I think this method is specific to phase1, should we renamed to something like getPhase1OnlineTest of getPhase1Test?

app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -279,51 +278,38 @@ trait OnlineTestService {
...
-  }
 
   private[services] def getAdjustedTime(minimum: Int, maximum: Int, percentageToIncrease: Int) = {
     val adjustedValue = math.ceil(minimum.toDouble * (1 + percentageToIncrease / 100.0))
     math.min(adjustedValue, maximum).toInt
   }
 
   private[services] def buildInviteApplication(application: OnlineTestApplication, token: String, userId: Int, scheduleId: Int) = {
-    val scheduleCompletionUrl = s"${gatewayConfig.candidateAppUrl}/fset-fast-stream/online-tests/complete/$token"
+    val scheduleCompletionUrl = s"${gatewayConfig.candidateAppUrl}/fset-fast-stream/online-tests/phase1"
```
mgilgar: I think we should be renamed to include Prefix like this scheduleCompletionUrlPrefix, scheduleCompletionBaseURL or something similar.

app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -279,51 +278,38 @@ trait OnlineTestService {
...
-      InviteApplicant(scheduleId, userId, scheduleCompletionUrl, resultsURL = None, timeAdjustments = None)
+      InviteApplicant(scheduleId,
+        userId,
+        s"$scheduleCompletionUrl/complete/$token",
+        resultsURL = None
+      )
     } else {
-      val timeAdjustments = getTimeAdjustments(application)
-      InviteApplicant(scheduleId, userId, scheduleCompletionUrl, resultsURL = None, timeAdjustments)
+      val completionUrl = if (scheduleIdByName("sjq") == scheduleId) {
```
mgilgar: If you rename the other one, this could be scheduleCompletionUrl

test/services/onlinetesting/OnlineTestServiceSpec.scala
```diff
@@ -241,10 +214,10 @@ class OnlineTestServiceSpec extends PlaySpec with BeforeAndAfterEach with Mockit
           .thenReturn(Future.successful(registration))
         when(cubiksGatewayClientMock.inviteApplicant(any[InviteApplicant])(any[HeaderCarrier]))
           .thenReturn(Future.successful(invitation))
-        when(cdRepositoryMock.find(UserId))
+        when(cdRepositoryMock.find("userId"))
```
mgilgar: "userId" this can be a val.

test/services/onlinetesting/OnlineTestServiceSpec.scala
```diff
@@ -277,45 +252,49 @@ class OnlineTestServiceSpec extends PlaySpec with BeforeAndAfterEach with Mockit
           .thenReturn(Future.successful(registration))
         when(cubiksGatewayClientMock.inviteApplicant(any[InviteApplicant])(any[HeaderCarrier]))
           .thenReturn(Future.successful(invitation))
-        when(cdRepositoryMock.find(UserId)).thenReturn(Future.successful(contactDetails))
-        when(emailClientMock.sendOnlineTestInvitation(eqTo(EmailContactDetails), eqTo(PreferredName), eqTo(ExpirationDate))(
-          any[HeaderCarrier]
-        )).thenReturn(Future.successful(()))
+        when(cdRepositoryMock.find("userId")).thenReturn(Future.successful(contactDetails))
```
mgilgar: Some comment, probably "userId" can be a val.

test/services/onlinetesting/OnlineTestServiceSpec.scala
```diff
@@ -365,34 +328,44 @@ class OnlineTestServiceSpec extends PlaySpec with BeforeAndAfterEach with Mockit
...
 
   "build invite application" should {
-    "return an InviteApplication with no time adjustments if gis and application has no time adjustments" in new OnlineTest {
-      onlineTestService.buildInviteApplication(applicationForOnlineTestingGisWithNoTimeAdjustments,
-        "", CubiksUserId, ScheduleId) must be(inviteApplicantGisWithNoTimeAdjustments)
-    }
+    "return an InviteApplication for a GIS candidate" in new OnlineTest {
+      val result = onlineTestService.buildInviteApplication(
+        onlineTestApplication.copy(guaranteedInterview = true),
+        "token", cubiksUserId, sjqScheduleId
```
mgilgar: Maybe "token" can be a val, if it is going to be reused later.

mgilgar (commented):
Good work although I have made some comments mostly about names.

mgilgar:
There are some problems in travis with this project.

app/controllers/OnlineTestController.scala
benjstephenson: Agreed - we will go back and clean up these things once we have released as there are a few areas like this we need to tidy up but we need to agree as a team what the naming conventions will be.

mgilgar: ok

PR #56, opened 2016-09-22
FSET-697: Candidate progress report service implemented

satishkamatam:


app/model/CandidateProgressReport.scala
```diff
@@ -0,0 +1,29 @@
+/*
```
mgilgar: Not sure if we do it before but I think it might be a good idea to put all report case classes in its own package. The reason is they usually agregrate information from different classes. What do you think?

app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -374,7 +376,7 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService)(implic
 
   }
 
-  override def overallReportNotWithdrawn(frameworkId: String): Future[List[Report]] =
+  override def overallReportNotWithdrawn(frameworkId: String): Future[List[CandidateProgressReport]] =
```
mgilgar: If the report is named "Candidate Progress Report" I think is a good idea to rename the method to something like candidateProgressReportNotWithdrawn.

app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -386,35 +388,53 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService)(implic
       BSONDocument("applicationStatus" -> BSONDocument("$ne" -> "WITHDRAWN"))
     )))
 
-  override def overallReport(frameworkId: String): Future[List[Report]] =
+  override def overallReport(frameworkId: String): Future[List[CandidateProgressReport]] =
```
mgilgar: If the report is named "Candidate Progress Report" I think is a good idea to rename the method to something like candidateProgressReport.

app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -386,35 +388,53 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService)(implic
...
     overallReport(BSONDocument("frameworkId" -> frameworkId))
 
-  private def overallReport(query: BSONDocument): Future[List[Report]] = {
+  private def overallReport(query: BSONDocument): Future[List[CandidateProgressReport]] = {
```
mgilgar: If the report is named "Candidate Progress Report" I think is a good idea to rename the method to something like candidateProgressReport.

app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -216,11 +218,11 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService)(implic
     val projection = BSONDocument("applicationStatus" -> 1, "progress-status-dates" -> 1, "_id" -> 0)
 
     collection.find(query, projection).one[BSONDocument] map {
-      case Some(document) => {
+      case Some(document) =>
         val status = document.getAs[String]("applicationStatus").get
```
mgilgar: This is old code but  think should be renamed to applicationStatus

app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -216,11 +218,11 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService)(implic
...
         val statusDate = document.getAs[BSONDocument]("progress-status-dates").flatMap(_.getAs[LocalDate](status.toLowerCase))
```
mgilgar: This is old code but  think should be renamed to progressStatusDate

app/model/CandidateProgressReport.scala
satishkamatam: yes. moved to report package

app/repositories/application/GeneralApplicationRepository.scala
satishkamatam: renamed

app/repositories/application/GeneralApplicationRepository.scala
satishkamatam: renamed to progressStatusDate

app/repositories/application/GeneralApplicationRepository.scala
satishkamatam: renamed to candidateProgressReportNotWithdrawn

app/repositories/application/GeneralApplicationRepository.scala
satishkamatam: renamed to candidateProgressReport

app/repositories/application/GeneralApplicationRepository.scala
satishkamatam: renamed to candidateProgressReport

PR #57, opened 2016-09-22
FSET 642: Extend a candidates online tests

henricook:
Depends on: https://github.tools.tax.service.gov.uk/HMRC/fset-faststream-admin-frontend/pull/16


app/controllers/OnlineTestController.scala
```diff
@@ -62,10 +68,10 @@ trait OnlineTestController extends BaseController {
 
   import Commands.Implicits._
 
-  def getOnlineTest(userId: String) = Action.async { implicit request =>
-    onlineTestingService.getPhase1TestProfile(userId).map {
+  def getOnlineTest(applicationId: String) = Action.async { implicit request =>
```
mgilgar: I guess this method could be renamed to getPhase1OnlineTests or getPhase1Tests. I made the same suggestion to ben for this method in another review.

app/controllers/Phase1TestGroupController.scala
```diff
@@ -17,37 +17,29 @@
...
+  val phase1TestExtensionService: OnlineTestExtensionService
 
-  def getStatus(applicationId: String) = Action.async { implicit request =>
-    for {
-      phase1TestGroupOpt <- phase1TestRepository.getPhase1TestProfile(applicationId)
-    } yield {
-      phase1TestGroupOpt.map { phase1TestGroup => Ok(Json.toJson(phase1TestGroup)) }.getOrElse(NotFound)
+  def extend(applicationId: String): Action[JsValue] = Action.async(parse.json) { implicit request =>
+    withJsonBody[OnlineTestExtension] { extension =>
+      phase1TestExtensionService.extendTestGroupExpiryTime("PHASE1", applicationId, extension.extraDays).map(_ => ())
```
mgilgar: I think the name of the service (val) should be renamed to something like onlineTestExtensionService, if you include phase1 it feels specific but then when you call the method you are passing "PHASE1" which would not be necessary if it was actually a phase1TestExtensionService.

app/repositories/application/OnlineTestRepository.scala
```diff
@@ -142,7 +135,16 @@ class OnlineTestMongoRepository(dateTime: DateTimeFactory)(implicit mongo: () =>
...
     ))
 
-    collection.update(query, applicationStatusBSON, upsert = false) map ( _ => () )
+    Logger.warn(s" ========== Running with app id: $applicationId and $phase1TestProfile")
+
+    collection.update(query, applicationStatusBSON, upsert = false) map { status =>
+      if (status.n != 1) {
+        val msg = s"${status.n} rows affected when inserting or updating instead of 1! (App Id: $applicationId)"
+        Logger.warn(msg)
+        throw UnexpectedException(msg)
```
mgilgar: UnexpectedException feels a bit unspecific.

app/services/onlinetesting/OnlineTestExpiryService.scala
```diff
@@ -65,7 +66,7 @@ class OnlineTestExpiryServiceImpl(
     }
 
   override def commitExpiredProgressStatus(expiringTest: ExpiringOnlineTest): Future[Unit] =
```
mgilgar: If this is for phase1, maybe we should include that in the name of the method, i.e. commitPhase1TestsExpiredProgressStatus

app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -328,9 +325,11 @@ trait OnlineTestService {
   }
 
   def markAsStarted(cubiksUserId: Int): Future[Unit] = {
```
mgilgar: I think "markAsStarted" that should include phase1 in the name, maybe "markAsPhase1Started"

app/services/testdata/Phase1TestsInvitedStatusGenerator.scala
```diff
@@ -42,22 +43,20 @@ trait Phase1TestsInvitedStatusGenerator extends ConstructiveGenerator {
   def generate(generationId: Int, generatorConfig: GeneratorConfig)(implicit hc: HeaderCarrier) = {
 
     val sjqTest = Phase1Test(
-      cubiksUserId = 117344,
+      cubiksUserId = scala.util.Random.nextInt(10000000),
```
mgilgar: Don't we need to refer to a real cubiksUserId?
 It might be useful to create a user at that status and then try to pass the test (i.e. to verify the integration with cubiks screens is right), if the cubiks user id is ramdon, we can't.

app/controllers/OnlineTestController.scala
henricook: Yep, if it's ok we'll log this for tidy up afterwards - now that this is working and needed relatively urgently i'd rather not touch it if i don't have to ;-)

app/controllers/Phase1TestGroupController.scala
henricook: Groupkey wasn't being used, I was building something generic but the existing trend seemed to be towards phase1-specific stuff - i've now removed this parameter

app/controllers/OnlineTestController.scala
mgilgar: Ok, that's fine

app/controllers/Phase1TestGroupController.scala
mgilgar: ok

app/services/testdata/Phase1TestsInvitedStatusGenerator.scala
henricook: At the moment:

- I don't think we test any state transitions (i.e. nothing that would use real cubiks) - we could aim to start doing this eventually which would require proper cubiks user ids
- I don't particularly want the network latency of talking to cubiks in user generation right now

... in this case it doesn't need to be real because we're generating fake test data with the fake id, and using the fake id throughout generation - we never contact actual cubiks for anything. So we get away with it :-)

PR #58, opened 2016-09-22
FSET-638: Reminder to complete the online exercises

Lansalo:


it/repositories/OnlineTestRepositorySpec.scala
```diff
@@ -198,7 +197,77 @@ class OnlineTestRepositorySpec extends MongoRepositorySpec {
...
+      }
+    }
+
+    "return no results" when {
+      val date = DateTime.now().plusHours(22)
+      val testProfile = Phase1TestProfile(expirationDate = date, tests = List(phase1Test))
+      "there are no application in PHASE1_TESTS" in {
+        createApplicationWithAllFields(UserId, AppId, "frameworkId", "SUBMITTED")
+        onlineTestRepo.insertOrUpdatePhase1TestGroup(AppId, testProfile).futureValue
+        updateApplication(BSONDocument("applicationStatus" -> ApplicationStatus.IN_PROGRESS))
```
ollyw: It looks like UpdateApplication is not awaited upon, so there is a race condition in this test and others that us updateApplication. Perhaps change the update application def to use .futurevalue in its implementation

app/model/ReminderNotice.scala
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package model
+
+import model.ProgressStatuses.{ PHASE1_TESTS_FIRST_REMINDER, PHASE1_TESTS_SECOND_REMINDER, ProgressStatus }
+
+final case class ReminderNotice(hoursBeforeReminder: Int, progressStatuses: ProgressStatus) {
+  require(hoursBeforeReminder > 0)
```
ollyw: I am liking the use of require

app/services/onlinetesting/OnlineTestExpiryService.scala
```diff
@@ -45,13 +52,28 @@ class OnlineTestExpiryServiceImpl(
...
     otRepository.nextExpiringApplication.flatMap {
       case Some(expiredTest) => processExpiredTest(expiredTest)
       case None => Future.successful(())
     }
   }
 
+  override def processReminder(expiringTest: NotificationExpiringOnlineTest, reminder: ReminderNotice): Future[Unit] =
+    for {
+      emailAddress <- candidateEmailAddress(expiringTest.userId)
+      //_ <- commitExpiredProgressStatus(expiringTest)
```
ollyw: Is this still required>

Lansalo: Yes, it is but I'll have to work on it and change the behavior.

it/repositories/OnlineTestRepositorySpec.scala
Lansalo: I had a look at it and it shouldn't but I'll double check before pushing the next patch ;-)
Perhaps I can make it more explicit as from what i can remember it was waiting  inside the createApplicationWithAllFields for the future to complete which is not probably really clear from the method's signature.

app/model/ReminderNotice.scala
henricook: Me too, although the exceptions thrown when a message isn't specified are remarkably vague. I'd suggest we enforce that a message must always be set, for example here:

require(hoursBeforeReminder > 0, "Hours before reminder was negative")

... something liked that so that once we are using a lot of these it's easier to track down which one has thrown up, and perhaps give people an informative enough error message to hazard a fix without having to dig into the code

app/repositories/application/OnlineTestRepository.scala
```diff
@@ -163,6 +165,24 @@ class OnlineTestMongoRepository(dateTime: DateTimeFactory)(implicit mongo: () =>
...
+    val query = BSONDocument("$and" -> BSONArray(
+      BSONDocument(
+        "applicationStatus" -> ApplicationStatus.PHASE1_TESTS
+      ),
+      BSONDocument(
+        "testGroups.PHASE1.expirationDate" -> BSONDocument(
+          "$lte" -> dateTime.nowLocalTimeZone.plusHours(reminder.hoursBeforeReminder)) // Serialises to UTC.
+      ),
+      BSONDocument("$and" -> BSONArray(
+        BSONDocument("progress-status.PHASE1_TESTS_COMPLETED" -> BSONDocument("$ne" -> true)),
```
henricook: This should use the case object's from the ProgressStatuses class please, rather than strings

henricook:
There's something very similar to the TimeUnit's you've used in the scala.concurrent.duration package - http://www.scala-lang.org/api/2.9.3/scala/concurrent/duration/package.html

... however we can still merge this, it's only a small duplication at the moment

app/connectors/EmailClient.scala
```diff
@@ -98,12 +114,17 @@ trait CSREmailClient extends EmailClient {
       Map("name" -> name)
     )
   }
+
+
```
henricook: Minor point: evul spaces

app/connectors/EmailClient.scala
```diff
@@ -126,4 +147,9 @@ object EmailDateFormatter {
     dateTime.toString("d MMMM yyyy, h:mma")
       .replace("AM", "am").replace("PM", "pm") // Joda time has no easy way to change the case of AM/PM
   }
+
+  def convertToHoursOrDays(timeUnit: TimeUnit, timeLeft: Int): String = {
```
henricook: So timeLeft is always in hours? And this method converts? It was tough to understand, could you perhaps change the name of the 'timeLeft' parameter to 'timeLeftInHours'?

app/services/onlinetesting/OnlineTestExpiryService.scala
```diff
@@ -46,29 +46,67 @@ class OnlineTestExpiryServiceImpl(
...
-  override def processExpiredTest(expiringTest: ExpiringOnlineTest): Future[Unit] =
+  override def commitExpiredProgressStatus(expiringTest: ExpiringOnlineTest): Future[Unit] =
+    applicationRepository.addProgressStatusAndUpdateAppStatus(expiringTest.applicationId, PHASE1_TESTS_EXPIRED).map { _ =>
+      audit("ExpiredOnlineTest", expiringTest)
+    }
+
+  private def processReminder(expiringTest: NotificationExpiringOnlineTest, reminder: ReminderNotice): Future[Unit] =
+    for {
+      emailAddress <- candidateEmailAddress(expiringTest.userId)
+      _ <- emailCandidateForExpiringTestReminder(expiringTest, emailAddress, reminder)
```
henricook: I think there was a small conversation last week about always emailing after database activity.

With this ordering if for some reason the database write files we'll end up spamming the user (as the scheduler will re-run, send the email, fail to write to the DB again).

If the email is last the failure to write to the DB will stop the email sending and the scheduler will eventually recover

app/model/ReminderNotice.scala
Lansalo: Agree and done :-)

app/repositories/application/OnlineTestRepository.scala
Lansalo: Done

Lansalo: Done

app/connectors/EmailClient.scala
Lansalo: Done

app/connectors/EmailClient.scala
Lansalo: Done

app/services/onlinetesting/OnlineTestExpiryService.scala
Lansalo: Done

henricook:
Ah you used the scala.concurrent.duration stuff - nice! :+1:

PR #59, opened 2016-09-22
Remove test progress statuses on reset

lukastymo:


app/repositories/application/OnlineTestRepository.scala
```diff
@@ -184,6 +186,21 @@ class OnlineTestMongoRepository(dateTime: DateTimeFactory)(implicit mongo: () =>
     collection.update(query, applicationStatusBSON, upsert = false) map ( _ => () )
   }
 
+  override def removeProgressStatuses(appId: String, progressStatuses: List[ProgressStatus]): Future[Unit] = {
```
benjstephenson: If this method only works for PHASE1 statuses then I think we should make this clear in the method name.

app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -112,6 +111,17 @@ trait OnlineTestService {
     registerAndInviteForTestGroup(application, getScheduleNamesForApplication(application))
   }
 
+  def resetPhase1Tests(application: OnlineTestApplication, testNamesToRemove: List[String]): Future[Unit] = {
+    for {
+    - <- registerAndInviteForTestGroup(application, testNamesToRemove)
+    } yield {
+      auditService.logEventNoRequest(
+        "Phase1TestsReseted",
```
benjstephenson: "Phase1TestsReset"

app/repositories/application/OnlineTestRepository.scala
lukastymo: I thought that the whole repo is for Phase1 (bad naming). I like small repos. We have already few methods only for phase1.

lukastymo: Anyway it's worth to keep same naming. I'll renam to phase1

PR #60, opened 2016-09-23
Fix the check for a fastpass candidate in nextCandidateReadyForOnlineTesting

benjstephenson:
Fix the check for a fastpass candidate in nextCandidateReadyForOnlineTesting.  We should be using the
"fastPassReceived" flag.


ollyw:
Looks harmless enough. Let's see if the tests pass

PR #68, opened 2016-09-26
Escape potentially exploitable characters before they are injected into a regex

henricook:


ollyw:
Looks good! Once the build passes, this is going in

PR #69, opened 2016-09-26
Return 404 for Cubiks data missing

ollyw:


benjstephenson:
Should we update the CannotFindTestByCubiksId to extend a NotFoundException so that it's inline with the model in fset-faststream?

PR #70, opened 2016-09-26
FSET-782: Update school model to reflect the CSV

lukastymo:


ollyw:
@lukastymo you wrote tests even under pressure 💯

PR #72, opened 2016-09-27
Spike: Track all events in the system

lukastymo:


lukastymo:
Main concepts:
- JSON is sent to Controller, Controller is creating Commands (operations which trigger state update)
- Commands are passed to Business Services
- **Business Services "do the job" and return List of Events as a response**
- List of events are passed to Event Service
- Event Service processes the events and choose Event Handlers based on event types
- 3 types of Events and EventHandlers: AuditEventHandler (wrapper around AuditService), MongoEventHandler (new handler required by the story: FSET-681), EmailEventHandler (we may treat sending the email also as events - this point I'm not sure, but it demonstrates  extensibility)

PR #76, opened 2016-09-29
FSET-833: Remind candidates with extensions to complete the online ex…

Lansalo:
…ercises

### DIFF
--- a/test/services/onlinetesting/OnlineTestExtensionServiceSpec.scala
+++ b/test/services/onlinetesting/OnlineTestExtensionServiceSpec.scala
@@ -68,12 +68,10 @@ class OnlineTestExtensionServiceSpec extends PlaySpec with ScalaFutures with Moc
 
         underTest.extendTestGroupExpiryTime(applicationId, extraDays).futureValue(timeout(Span(1, Seconds))) mustBe (())
 
-        verify(mockAppRepository).findProgress(eqTo(applicationId))
-        verify(mockOtRepository).getPhase1TestGroup(eqTo(applicationId))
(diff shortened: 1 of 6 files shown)


test/services/onlinetesting/OnlineTestExtensionServiceSpec.scala
```diff
@@ -14,66 +14,166 @@
...
+        when(mockAppRepository.findProgress(any())).thenReturnAsync(successfulProgressResponse)
+        when(mockOtRepository.getPhase1TestGroup(applicationId)).thenReturnAsync(None)
+        whenReady(underTest.extendTestGroupExpiryTime(applicationId, twoExtraDays).failed) { e =>
+          e mustBe (noTestProfileFoundError)
+        }
+        verify(mockAppRepository).findProgress(eqTo(applicationId))
+        verify(mockOtRepository).getPhase1TestGroup(eqTo(applicationId))
+        verifyNoMoreInteractions(mockAppRepository, mockOtRepository, mockAuditService, mockDateFactory)
+      }
+      "remove status return an error and no audit event is emitted" in new TestFixture {
```
Lansalo: "remove status return an error and no audit event is emitted" is the behavior of the previous version so this test just check that, but is it really what we want in this unlikely scenario?

henricook: I think it's OK, audit events should only happen when state is changed - although we should probably emit a normal log to say something's gone wrong

henricook:
Nice refactoring!

PR #77, opened 2016-09-30
Space out calls to register/invite to Cubiks

benjstephenson:


app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -133,9 +136,12 @@ trait OnlineTestService extends ResetPhase1Test {
     // going on in the background.
     // The approach to fixing it here is to generate futures that return Try[A] and then all futures will be
     // traversed. Afterward, we look at the results and clear up the mess
+    // TODO we space out calls to Cubiks because it appears they fail when they are too close together.
```
henricook: It's a horrendously minor point, but this isn't really a TODO. One day i hope to search all the TODOs and do something about them, but this doesn't need anything doing about it, i don't think?

PR #80, opened 2016-10-02
Online test tidyup

benjstephenson:
Refactor the online test repository into a common trait for use with phase 1 test repository and then for other tests (e-tray, video interview etc).
The Phase1TestProfile case class and other models for test profiles should extend a common TestProfile class.

I'd like to revisit the OnlineTestService to pull out all the phase1 specific stuff.
I'm pretty confident a lot of what we're doing with tests can be generic enough to be reused.


ollyw:
This is looking good! Presumably it is going to break compatibility with the cubiks gateway, so we had better make sure they get merged/deployed together

benjstephenson:
We should still be compatible with the Cubiks gateway - the controllers/contracts haven't been changed.  Yet :P

app/model/OnlineTestCommands.scala
```diff
@@ -25,42 +25,46 @@ import reactivemongo.bson.{BSONDocument, BSONHandler, Macros}
...
   }
 
-  case class Phase1TestProfile(
-                                expirationDate: DateTime,
-                                tests: List[Phase1Test]
-  ) {
-    def activeTests = tests filter (_.usedForResults)
+  trait TestProfile {
+    def expirationDate: DateTime
+    def tests: List[Phase1Test]
```
Lansalo: Is this list meant to contain only Phase1Test? If so I don't see much benefit in using this trait but perhaps I'm missing a bit of background here.

app/model/OnlineTestCommands.scala
```diff
@@ -25,42 +25,46 @@ import reactivemongo.bson.{BSONDocument, BSONHandler, Macros}
...
+    def hasNotStartedYet = activeTests.forall(_.startedDateTime.isEmpty)
     def hasNotCompletedYet =  activeTests.exists(_.completedDateTime.isEmpty)
-
     def hasNotResultReadyToDownloadForAllTestsYet =  activeTests.exists(!_.resultsReadyToDownload)
   }
 
   object Phase1TestProfile {
     import repositories.BSONDateTimeHandler
-    implicit val phase1TestProfileHandler: BSONHandler[BSONDocument, Phase1TestProfile] = Macros.handler[Phase1TestProfile]
+    implicit def bsonHandler: BSONHandler[BSONDocument, Phase1TestProfile] = Macros.handler[Phase1TestProfile]
```
Lansalo: Why from val to def?

app/model/OnlineTestCommands.scala
benjstephenson: Once we start the phase2 tests work we should know more about what a Phase2Test case class would look like and then create a generic Test trait to use in this list.

app/model/OnlineTestCommands.scala
benjstephenson: Not sure, I'll change it back.  Think it was when I was playing around with the function types in the OnlineTestRepo trait.

app/model/OnlineTestCommands.scala
Lansalo: Makes sense.

Lansalo:
Looks fine to me. Let's wait for the builds to complete and if Olli is happy as well I think we can merge this.

ollyw:
if @Lansalo is happy, I am happy. Merge at will

PR #81, opened 2016-10-04
Add the 'faststream' service name to the download report url.

benjstephenson:


henricook:
Tests pass and it's in

PR #82, opened 2016-10-04
FSET-696: Evaluate application in Phase1 tests - Next application

lukastymo:


it/repositories/onlinetesting/Phase1EvaluationMongoRepositorySpec.scala
```diff
@@ -0,0 +1,112 @@
...
+
+      result mustBe Some(ApplicationToPhase1Evaluation(
+        "app1",
+        isGis = true,
+        Phase1TestProfile(now, testsWithResult),
+        selectedSchemes))
+    }
+  }
+
+  private def insertApp(appId: String, applicationStatus: ApplicationStatus, tests: Option[List[Phase1Test]] = None,
```
benjstephenson: Not really a review comment, but so many test specs have this method to some varying degree, someone should pull them all out into a helper class.

app/repositories/onlinetesting/Phase1EvaluationRepository.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  def nextApplicationReadyForPhase1ResultEvaluation: Future[Option[ApplicationToPhase1Evaluation]] = {
+    // TODO: Add BSONDocument("passmarkEvaluation.passmarkVersion" -> BSONDocument("$exists" -> false))
+    val query = BSONDocument("$or" -> BSONArray(
+      BSONDocument("$and" -> BSONArray(
+        BSONDocument("applicationStatus" -> ApplicationStatus.PHASE1_TESTS),
+        BSONDocument(s"progress-status.${ProgressStatuses.PHASE1_TESTS_RESULTS_RECEIVED}" -> true)
+      )),
+      BSONDocument("$and" -> BSONArray(
+        BSONDocument("applicationStatus" -> ApplicationStatus.PHASE2_TESTS),
```
lukastymo: It's called "Business Logic" :)

The case is that they can be promoted to Phase2, but they may have some ambers (not evaluated schemes). In Phase2 evaluation not only check phase2 results, but also what was the result per scheme in Phase1

PR #101, opened 2016-10-12
Fset 727

Lansalo:
This is a refactoring needed in order to accommodate the new scheduler (reminders and expiry) for e-tray. I decided to push it in it's own PR so can be tested and make sure that doesn't break anything of what we have currently in place. Some further refactoring might be needed in the area of the service but that will be left to another story.


Lansalo:
This should be merged along with: https://github.tools.tax.service.gov.uk/HMRC/fset-faststream-email/pull/8

PR #107, opened 2016-10-13
OnlineTestResults STC event

benjstephenson:


app/services/onlinetesting/Phase2TestService.scala
```diff
@@ -123,8 +125,11 @@ trait Phase2TestService extends OnlineTestService {
       invitedApplicants <- inviteApplicants(registeredApplicants)
       _ <- insertPhase2TestGroups(invitedApplicants)(invitationDate, expirationDate)
       _ <- emailInviteToApplicants(candidatesToProcess)(hc, invitationDate, expirationDate)
-    } yield candidatesToProcess.foreach { candidate =>
-      audit("Phase2TestInvitationProcessComplete", candidate.userId)
+    } yield eventSink {
+      Future.successful(candidatesToProcess.map { candidate =>
```
ollyw: Future[Future[Unit]] alert. We need to do some `} flatmap identity` action

app/services/.#AuditService.scala
```diff
@@ -0,0 +1 @@
+bste@Benjis-MBP.30300
```
ollyw: Perhaps we need another .gitignore setting for spacemacs?

PR #113, opened 2016-10-14
Fset 677 - Admin withdraw a candidate from Fast Stream

kjharkin:


ollyw:
No review comments!

app/services/application/ApplicationService.scala
```diff
@@ -31,18 +36,41 @@ import scala.concurrent.{ ExecutionContext, Future }
...
-      AuditEvents.ApplicationWithdrawn(Map("applicationId" -> applicationId, "withdrawRequest" -> withdrawRequest.toString)) ::
-      Nil
-    }
+    (implicit hc: HeaderCarrier, rh: RequestHeader): Future[Unit] = {
+
+    appRepository.find(applicationId).flatMap{
+      case Some(candidate) =>
+        for {
+          cd <- cdRepository.find(candidate.userId)
+        } yield {
```
ollyw: Yielding a future in this yield block causes a Future[Future[Unit] to be returned. Once this gets to `appRepository.find(applicationId).flatMap{`, it will be mapped onto another future, but the flatmapping only flattens one level, so you still end up with Future[Future[Unit]]. Sad times. This means that failures will not necessarily be caught

ollyw: Yielding a future in this yield block causes a Future[Future[Unit] to be returned. Once this gets to `appRepository.find(applicationId).flatMap{`, it will be mapped onto another future, but the flatmapping only flattens one level, so you still end up with Future[Future[Unit]]. Sad times. This means that failures will not necessarily be caught

ollyw:
I changed my mind after finding the Future issue

kjharkin:
Devastated.

ollyw:
LGTM 👍

PR #115, opened 2016-10-14
Phase2 test data generator

benjstephenson:


app/services/testdata/Phase2TestsInvitedStatusGenerator.scala
```diff
@@ -0,0 +1,73 @@
...
+}
+
+trait Phase2TestsInvitedStatusGenerator extends ConstructiveGenerator {
+  val otRepository: Phase2TestRepository
+  val gatewayConfig: CubiksGatewayConfig
+
+  def generate(generationId: Int, generatorConfig: GeneratorConfig)(implicit hc: HeaderCarrier, rh: RequestHeader) = {
+
+    val etray = CubiksTest(
+      cubiksUserId = scala.util.Random.nextInt(1000000000),
```
ollyw: This int will occasionally overflow. Suggest using Int.MaxValue. In theory, it should be Int.MaxValue + 1, but that just looks scruffy

app/services/testdata/StatusGeneratorFactory.scala
```diff
@@ -58,6 +58,14 @@ object StatusGeneratorFactory {
         }
       case (PHASE1_TESTS, Some(PHASE1_TESTS_COMPLETED)) => Phase1TestsCompletedStatusGenerator
       case (PHASE1_TESTS, Some(PHASE1_TESTS_RESULTS_RECEIVED)) => Phase1TestsResultsReceivedStatusGenerator
+      case (PHASE2_TESTS, Some(PHASE2_TESTS_INVITED)) => Phase2TestsInvitedStatusGenerator
+      case (PHASE2_TESTS, Some(PHASE2_TESTS_STARTED)) => Phase2TestsStartedStatusGenerator
+      case (PHASE2_TESTS, Some(PHASE2_TESTS_EXPIRED)) =>
+        if (generatorConfig.phase1StartTime.isDefined) {
```
ollyw: Seems strange that phase2 could be expired when phase1StartTime is not defined. I am guessing this is some form of hack just for tests though :-)

app/services/testdata/GeneratorConfig.scala
```diff
@@ -20,20 +20,22 @@ import model.EvaluationResults.Result
...
+  loc1scheme1Passmark: Option[Result],
+  loc1scheme2Passmark: Option[Result],
+  previousStatus: Option[String],
+  confirmedAllocation: Boolean = true,
+  dob: Option[LocalDate],
+  postCode: Option[String],
+  country: Option[String],
+  phase1StartTime: Option[DateTime],
+  phase1ExpiryTime: Option[DateTime],
+  tscore: Option[Double]
```
henricook: This should probably be phase1 or phase2 tscore?

henricook: Talked about this, could potentially be used for phase1 and phase2 (if generating a phase2 test phase1 is implied as passed etc)

app/services/testdata/Phase2TestsInvitedStatusGenerator.scala
```diff
@@ -0,0 +1,73 @@
...
+      tests = List(etray)
+    )
+
+    for {
+      candidateInPreviousStatus <- previousStatusGenerator.generate(generationId, generatorConfig)
+      _ <- otRepository.insertOrUpdateTestGroup(candidateInPreviousStatus.applicationId.get, phase2TestGroup)
+    } yield {
+      val etray = phase2TestGroup.tests.head
+
+      candidateInPreviousStatus.copy(phase1TestGroup = Some(
```
henricook: As you mentioned this needs changing from phase1 to phase2

app/services/testdata/StatusGeneratorFactory.scala
benjstephenson: Dirty hack to avoid going over the 22 attribute limit for the unapply method.  Rationale is that if you're generating a phase2 test then does it really matter when the phase1 was started as it should already be complete and passed.  We should probably redo this to pass in a JSON object or something.

PR #121, opened 2016-10-17
Fset 698 Download the diversity report

mgilgar:
https://jira.tools.tax.service.gov.uk/browse/FSET-698

This is part of a group or several PR that should be merged in the following order:
https://github.com/hmrc/fset-faststream/pull/121
https://github.com/hmrc/fset-faststream-frontend/pull/97
https://github.tools.tax.service.gov.uk/HMRC/fset-faststream-admin-frontend/pull/50
https://github.tools.tax.service.gov.uk/HMRC/fset-faststream-admin-acceptance-tests/pull/24


app/controllers/ReportingController.scala
```diff
@@ -46,24 +46,34 @@ trait ReportingController extends BaseController {
...
-      case None => NotFound
+  def diversityReport(frameworkId: String) = Action.async { implicit request =>
+    val reports = for {
+      applications <- appRepository.diversityReport(frameworkId)
+      questionnaires <- questionnaireRepository.findAllForDiversityReport
+      medias <- medRepository.findAll()
+    } yield {
+      for {
+        a <- applications
+        q <- Some(questionnaires.get(a.applicationId))
```
kjharkin: questionnnaires.get(a.applicationId) already returns an option. Why is this line not:

q = questionnaires.get(a.applicationId) ?

app/controllers/ReportingController.scala
```diff
@@ -46,24 +46,34 @@ trait ReportingController extends BaseController {
...
+        m <- Some(medias.get(a.userId))
```
kjharkin: medias.get(a.userId) already returns an option. Why is this line not:

m = medias.get(a.userId) ?

app/controllers/ReportingController.scala
```diff
@@ -83,6 +93,38 @@ trait ReportingController extends BaseController {
...
+  def createPreferencesAndContactReports(frameworkId: String) =
+    preferencesAndContactReports(nonSubmittedOnly = false)(frameworkId)
+
+  def applicationAndUserIdsReport(frameworkId: String) = Action.async { implicit request =>
+    appRepository.allApplicationAndUserIds(frameworkId).map { list =>
+      Ok(Json.toJson(list))
+    }
+  }
+
+  /*
```
mgilgar: I decide to comment what it is not used right now because it was very confusing for me a file that long, commenting it helped me alot. I am a bit scared of removing it completely, although I should. I check with BAs and they are not sure what report we will have in the future.

app/controllers/ReportingController.scala
mgilgar: done

app/repositories/QuestionnaireRepository.scala
```diff
@@ -35,7 +35,8 @@ import scala.language.postfixOps
 trait QuestionnaireRepository {
   def addQuestions(applicationId: String, questions: List[PersistedQuestion]): Future[Unit]
   def findQuestions(applicationId: String): Future[Map[String, String]]
-  def onlineTestPassMarkReport: Future[Map[String, PassMarkReportQuestionnaireData]]
+  def findForOnlineTestPassMarkReport: Future[Map[String, QuestionnaireReportItem]]
+  def findAllForDiversityReport: Future[Map[String, QuestionnaireReportItem]]
```
kjharkin: This is a subjective point but I don't like names like "findAllForDiversityReport". Surely, the repository is being asked to 'findAll' - what does it care what it's for.

mgilgar: The problem I was trying to solve is that both findForOnlineTestPassMarkReport and findAllForDiversityReport return the same kind of objects but in one of them we do not filter, we return everything and in the other one we filter the results. I thought about having only one method because in findForOnlineTestPassMarkReport we probably don't need the following lines:

val firstEmploymentQuestion = "When you were 14, what kind of work did your highest-earning parent or guardian do?"
      val query = BSONDocument(s"questions.$firstEmploymentQuestion" -> BSONDocument("$exists" -> BSONBoolean(true)))

So then, both methods are equal.

I think that filter is just an optimisation, in that case we can call it findAll, although even in that case I would prefer findAllAsReportItem and I would reserve findAll to a method that returns the persisted questionnaire object.

app/repositories/application/GeneralApplicationRepoBSONToModelHelper.scala
```diff
@@ -0,0 +1,238 @@
...
+import model.command.ProgressResponse
+import model.persisted.{ApplicationForNotification, Phase1TestProfile}
+import model.report._
+import model.{CivilServiceExperienceType, InternshipType}
+import org.joda.time.{DateTime, LocalDate}
+import reactivemongo.bson.{BSONDocument, _}
+import repositories._
+
+
+object GeneralApplicationRepoBSONToModelHelper extends GeneralApplicationRepoBSONToModelHelper
```
kjharkin: (Thought): There seems to have been a lot of refactoring here, defining custom readers. Would it not have been possible to have used ReactiveMongo Macros to cut down on bolierplate readers?

ollyw:
This looks good and the refactors are sensible. I would have preferred that there was less commented out code, even if we may use it in the future - that is what source versioning is for! There are many refactors that are going to cause merge conflicts everywhere. Ideally all the BSON refactoring stuff  would have been done in a separate PR and merged down ASAP, instead of bundling it with this)
So, is there anything stopping this from being merged to master now @mgilgar @kjharkin ?

ollyw:
PS, glad to see the back of the old Akka driven diversity report! Well done 👍

app/repositories/application/GeneralApplicationRepoBSONToModelHelper.scala
mgilgar: Probably, I thought that was too much to achieve in a single story, there is already a lot of refactoring. Those methods were just copied from GeneralApplicationRepository to make the class smaller. Maybe the next time we do another report we can do it.

kjharkin:
Well I guess if you're happy @ollyw then okay. Will merge and move onto the next project in this PR.

PR #126, opened 2016-10-18
Refactor statuses into a more type safe and foolproof system

ollyw:


lukastymo:
PR moved here: https://github.com/hmrc/fset-faststream/pull/144

PR #128, opened 2016-10-18
Fset 687: Expire the candidate's e-tray assessment window

Lansalo:


ollyw:
The code is good to go, but let's hold off merging until some test coverage is added for the phase2 specific functionality (just to make sure it is sending the right email templates, etc)

app/services/onlinetesting/Phase1TestService.scala
```diff
@@ -126,6 +126,13 @@ trait Phase1TestService extends OnlineTestService with ResetPhase1Test {
     registerAndInviteForTestGroup(application, getScheduleNamesForApplication(application))
   }
 
+  override def processNextExpiredTest(expiryTest: TestExpirationEvent)(implicit hc: HeaderCarrier, rh: RequestHeader): Future[Unit] = {
```
mgilgar: Probably you can have a generic processNextExpiredTest in OnlineTestService, I guess the logic will be the same in all phases, the only change would be the repo and service that you call. If you like it, I think you can skip for now, and deal with it in another sprint.

app/services/onlinetesting/Phase2TestService.scala
```diff
@@ -86,7 +86,7 @@ trait Phase2TestService extends OnlineTestService with ScheduleSelector {
     phase2TestRepo.nextApplicationsReadyForOnlineTesting
   }
 
-  override def processNextTestForReminder(reminder: ReminderNotice)(implicit hc: HeaderCarrier): Future[Unit] = {
+  override def processNextTestForReminder(reminder: ReminderNotice)(implicit hc: HeaderCarrier, rh: RequestHeader): Future[Unit] = {
```
mgilgar: Probably this be made generic as the method processNextExpiredTest.

app/services/onlinetesting/Phase3TestService.scala
```diff
@@ -91,11 +93,11 @@ trait Phase3TestService extends OnlineTestService with ResetPhase3Test with Even
     } yield audit("Phase3TestInvitationProcessComplete", application.userId)
   }
 
-  override def processNextTestForReminder(reminder: model.ReminderNotice)(implicit hc: HeaderCarrier): Future[Unit] = ???
+  override def processNextTestForReminder(reminder: model.ReminderNotice)(implicit hc: HeaderCarrier, rh: RequestHeader): Future[Unit] = ???
```
mgilgar: Probably this can be generic

app/services/onlinetesting/Phase3TestService.scala
```diff
@@ -77,6 +77,8 @@ trait Phase3TestService extends OnlineTestService with ResetPhase3Test with Even
     registerAndInviteForTestGroup(application, getInterviewIdForApplication(application))
   }
 
+  override def processNextExpiredTest(expiryTest: TestExpirationEvent)(implicit hc: HeaderCarrier, rh: RequestHeader): Future[Unit] = ???
```
mgilgar: Probably this can be generic

app/services/onlinetesting/Phase1TestService.scala
Lansalo: I remember I gave this a thought but you can't know the repo if you place it in OnlineTestService as is the trait that collect all the common functionality regardless of the phase (and repo)

app/services/onlinetesting/Phase2TestService.scala
Lansalo: Unfortunately we need to know the repo (phase 1 or two)

app/services/testdata/Phase2TestsExpiredFromStartedStatusGenerator.scala
```diff
@@ -16,11 +16,12 @@
```
Lansalo: As discusses, I think what we have right now, supports the application status but not the progress status.

PR #129, opened 2016-10-18
Change mongo query for results ready scheduler.

benjstephenson:
Check the resultsReadyToDownload flag and whether we already have tests saved or not when selecting candidates for the results scheduler.
If we have all the results for active tests then set the progress status to PHASE1_TESTS_RESULTS_RECEIVED.

### DIFF
--- a/app/repositories/onlinetesting/Phase1TestRepository.scala
+++ b/app/repositories/onlinetesting/Phase1TestRepository.scala
@@ -158,12 +158,24 @@ class Phase1TestMongoRepository(dateTime: DateTimeFactory)(implicit mongo: () =>
   }
 
   def nextTestGroupWithReportReady: Future[Option[Phase1TestWithUserIds]] = {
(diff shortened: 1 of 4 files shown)


app/repositories/onlinetesting/Phase1TestRepository.scala
```diff
@@ -158,11 +159,22 @@ class Phase1TestMongoRepository(dateTime: DateTimeFactory)(implicit mongo: () =>
   }
 
   def nextTestGroupWithReportReady: Future[Option[Phase1TestWithUserIds]] = {
+
+    /*db.getCollection('application').find(
```
ollyw: What is this little chap doing still in the code?

benjstephenson: Oh dear, I'll get that deleted next time I merge.

PR #131, opened 2016-10-19
Fix etray invites adjustments

satishkamatam:


test/services/onlinetesting/Phase2TestServiceSpec.scala
```diff
@@ -110,7 +110,7 @@ class Phase2TestServiceSpec extends PlaySpec with MockitoSugar with ScalaFutures
       phase2TestService.verifyDataStoreEvents(1, "OnlineExerciseResultSent")
     }
 
-    "exclude adjsutments candidates from the invite" in new Phase2TestServiceFixture {
+    "exclude adjsutments candidates from the invite" ignore new Phase2TestServiceFixture {
```
benjstephenson: How come we're ignoring this test now?  Should we not fix it?

satishkamatam: filtering logic in now in Phase2TestMongoRepo not in the service. And test has been added for the same. I can remove this test instead of ignoring.

benjstephenson: If it's not relevant here then I think we should remove it.

PR #136, opened 2016-10-20
FSET-688: Extend e-Tray

lukastymo:


app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -129,3 +129,16 @@ trait OnlineTestService extends EventSink  {
     }
   }
 }
+
+trait TimeExtension {
+  val dateTimeFactory: DateTimeFactory
+
+  def extendTime(alreadyExpired: Boolean, previousExpirationDate: DateTime) = { extraDays: Int =>
```
Lansalo: I personally would add the return type to this but it's something that can be done later

PR #137, opened 2016-10-20
Tests for the batch size logic in RandomSelection

benjstephenson:
Fix unit future return in evaluate phase 1 service spec.

### DIFF
--- a/test/repositories/RandomSelectionSpec.scala
+++ b/test/repositories/RandomSelectionSpec.scala
@@ -22,10 +22,30 @@ import org.scalatestplus.play.PlaySpec
 class RandomSelectionSpec extends PlaySpec {
   "RandomSelection" should {
     "calculate the correct batch size and random offset" in {
-      RandomSelection.calculateBatchSize(1, 1) mustBe (0,1)
+
+      RandomSelection.calculateBatchSize(1, 1) mustBe ((0,1))
       val (offset, size) = RandomSelection.calculateBatchSize(1, 9)
-      offset mustBe (0 +- 8)
+      offset must (be >= (0) and be <= (8))
       size mustBe 1
+
+      1 to 200 foreach { _ =>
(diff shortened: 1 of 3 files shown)


test/repositories/RandomSelectionSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import org.scalatest._
+import org.scalatestplus.play.PlaySpec
+
+class RandomSelectionSpec extends PlaySpec {
+  "RandomSelection" should {
+    "calculate the correct batch size and random offset" in {
+      RandomSelection.calculateBatchSize(1, 1) mustBe (0,1)
+      val (offset, size) = RandomSelection.calculateBatchSize(1, 9)
+      offset mustBe (0 +- 8)
```
ollyw: This allows -8, which is not what you wanted. Suggest `offset must (be >= (0) and be <= (8))`
In addition, this test should be run multiple times, as you are only testing a single number. Suggest using the property testing trait. http://www.scalatest.org/user_guide/property_based_testing

test/repositories/RandomSelectionSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+  "RandomSelection" should {
+    "calculate the correct batch size and random offset" in {
+
+      RandomSelection.calculateBatchSize(1, 1) mustBe ((0,1))
+      val (offset, size) = RandomSelection.calculateBatchSize(1, 9)
+      offset must (be >= (0) and be <= (8))
+      size mustBe 1
+
+      1 to 200 foreach { _ =>
+        val numberOfDocs = scala.util.Random.nextInt(10000)
```
ollyw: Can we adjust this value to 1000? It should speed the test up dramatically, and I am not convinced that it lowers our confidence in the validity of the test

PR #138, opened 2016-10-20
FSET-686: Reset etray

satishkamatam:


ollyw:
@satishkamatam Sorry to say that there are merge conflicts. Probably my fault for not reviewing quick enough :-(

app/controllers/OnlineTestController.scala
```diff
@@ -97,11 +91,32 @@ trait OnlineTestController extends BaseController {
...
+      def reset(onlineTestApp: OnlineTestApplication, actionTriggeredBy: String) =
+        phase2TestService.resetTests(onlineTestApp, actionTriggeredBy)
+          .map { _ =>
+            Ok(Json.toJson(ResetPhase2TestStatus(success = true)))
+          }
+          .recover {
+            case _: ResetLimitExceededException =>
+              Ok(Json.toJson(ResetPhase2TestStatus(success = false, resetLimitExceeded = true)))
+            case _: CannotResetPhase2Tests =>
+              Ok(Json.toJson(ResetPhase2TestStatus(success = false)))
```
ollyw: Perhaps this should be a 404, given that the CannotResetPhase2Tests exception inherits from NotFound. Certainly returning a 200 is not very 'RESTFul'. I would argue that ResetLimitExceededException could be HTTP code 424 and CannotResetPhase2Tests a 404. That way you don't need to return a success flag. What do you think?

ollyw: Sorry, I meant 423, not 424!

PR #142, opened 2016-10-24
Fix cubiks callback race conditions

satishkamatam:


app/repositories/onlinetesting/OnlineTestRepository.scala
```diff
@@ -58,6 +59,58 @@ trait OnlineTestRepository[U <: Test, T <: TestProfile[U]] extends RandomSelecti
...
+  }
+
+  def markTestAsInactive(cubiksUserId: Int) = {
+    val update = BSONDocument("$set" -> BSONDocument(
+      s"testGroups.$phaseName.tests.$$.usedForResults" -> false
+    ))
+    findAndUpdateCubiksTest(cubiksUserId, update)
+  }
+
+  def insertCubiksTests[T <: CubiksTestProfile](applicationId: String, newTestProfile: T) = {
```
satishkamatam: fixed compiler warning

app/services/onlinetesting/Phase1TestService.scala
```diff
@@ -268,26 +268,27 @@ trait Phase1TestService extends OnlineTestService with ResetPhase1Test {
...
-        } else {
-          t
+  private def insertOrAppendNewTests(applicationId: String, currentProfile: Option[Phase1TestProfile],
+                    newProfile: Phase1TestProfile): Future[Phase1TestProfile] = {
+    (currentProfile match {
+      case None => phase1TestRepo.insertOrUpdateTestGroup(applicationId, newProfile)
+      case Some(profile) =>
+        val scheduleIdsToArchive = newProfile.tests.map(_.scheduleId)
+        val inactiveTests = profile.tests.filter(t => scheduleIdsToArchive.contains(t.scheduleId)).map(_.cubiksUserId)
+        Future.traverse(inactiveTests)(phase1TestRepo.markTestAsInactive).flatMap { _ =>
```
benjstephenson: Some of these tests could already be marked as inactive, should we filter these out so we can avoid making repo calls when we don't need to?

app/services/onlinetesting/Phase1TestService.scala
```diff
@@ -357,7 +356,7 @@ trait Phase1TestService extends OnlineTestService with ResetPhase1Test {
   }
 
   def markAsReportReadyToDownload(cubiksUserId: Int, reportReady: CubiksTestResultReady): Future[Unit] = {
-    updatePhase1Test(cubiksUserId, updateTestReportReady(_:CubiksTest, reportReady)).flatMap { updated =>
+    updatePhase1Test(cubiksUserId, phase1TestRepo.updateTestReportReady(_:Int, reportReady)).flatMap { updated =>
       val allResultReadyToDownload = updated.phase1TestProfile.activeTests forall (_.resultsReadyToDownload)
```
ollyw: This looks like race condition

app/services/onlinetesting/Phase2TestService.scala
```diff
@@ -269,7 +273,7 @@ trait Phase2TestService extends OnlineTestService with ScheduleSelector {
   }
 
   def markAsReportReadyToDownload(cubiksUserId: Int, reportReady: CubiksTestResultReady): Future[Unit] = {
-    updatePhase2Test(cubiksUserId, updateTestReportReady(_: CubiksTest, reportReady)).flatMap { updated =>
+    updatePhase2Test(cubiksUserId, phase2TestRepo.updateTestReportReady(_: Int, reportReady)).flatMap { updated =>
       if (updated.phase2TestGroup.activeTests forall (_.resultsReadyToDownload)) {
```
ollyw: This looks like a race condition

app/services/onlinetesting/Phase1TestService.scala
ollyw: Cancel that. I didn't see that updatePhase1Test reloads.

app/services/onlinetesting/Phase1TestService.scala
```diff
@@ -268,26 +268,27 @@ trait Phase1TestService extends OnlineTestService with ResetPhase1Test {
...
-          t
+  private def insertOrAppendNewTests(applicationId: String, currentProfile: Option[Phase1TestProfile],
+                    newProfile: Phase1TestProfile): Future[Phase1TestProfile] = {
+    (currentProfile match {
+      case None => phase1TestRepo.insertOrUpdateTestGroup(applicationId, newProfile)
+      case Some(profile) =>
+        val scheduleIdsToArchive = newProfile.tests.map(_.scheduleId)
+        val inactiveTests = profile.tests.filter(t => scheduleIdsToArchive.contains(t.scheduleId)).map(_.cubiksUserId)
+        Future.traverse(inactiveTests)(phase1TestRepo.markTestAsInactive).flatMap { _ =>
+          phase1TestRepo.insertCubiksTests(applicationId, newProfile)
```
benjstephenson: I think I'm being dense here: should we not just be passing the new test that we want to add to the tests array and not everything in the profile?  We've already updated the tests from __scheduleIdsToArchive__ so why do we need to pass them again to the repo to be pushed to the existing list in mongo?

satishkamatam: insertCubiksTests method will just read expirationDate & tests from the profile and updates the test group

benjstephenson: we're still doing a $push for the test array along with the date update though.

ollyw:
PR looks good. If the manual testing looks good, let's merge it

PR #148, opened 2016-10-25
Implement diagnostic report

ollyw:


app/controllers/DiagnosticReportController.scala
```diff
@@ -41,4 +41,13 @@ trait DiagnosticReportController extends BaseController {
       case e => NotFound
     }
   }
+
+  def getAllApplications() = Action.async { implicit request =>
```
Lansalo: Same as above. Return type

test/scheduler/onlinetesting/EvaluatePhase1ResultJobSpec.scala
```diff
@@ -36,15 +36,15 @@ class EvaluatePhase1ResultJobSpec extends PlaySpec with MockitoSugar with ScalaF
 
   "Scheduler execution" should {
     "evaluate applications ready for evaluation" in new TestFixture {
-      when(mockEvaluateService.nextCandidatesReadyForEvaluation(any[Int])).thenReturn(Future.successful(Some(apps.toList, passmark)))
+      when(mockEvaluateService.nextCandidatesReadyForEvaluation(any[Int])).thenReturn(Future.successful(Some((apps.toList, passmark))))
```
Lansalo: Do we need those extra parenthesis?

ollyw: @Lansalo Yes. The signature is ```Future[Option[(List[ApplicationPhase1ReadyForEvaluation],Phase1PassMarkSettings)]]```,
 i.e a tuple. Not putting a tuple in compiles, because the compiler infers a tuple, but it generates a warning, and should not be done

ollyw: Yes, see previous comment

app/controllers/DiagnosticReportController.scala
ollyw: Interesting point. It is inconsistent in some ways to put a return type. I notice that none of the controllers in fset-faststream have return types. In many ways, there is a reduced value as most of the return types have to be Action[AnyContent]

app/controllers/DiagnosticReportController.scala
```diff
@@ -32,7 +32,7 @@ trait DiagnosticReportController extends BaseController {
 
   val drRepository: DiagnosticReportingRepository
 
-  def getUserById(userId: String) = Action.async { implicit request =>
+  def getApplicationById(userId: String) = Action.async { implicit request =>
```
ollyw: Good point. I'll change it

PR #150, opened 2016-10-26
FSET-936 Stop expired candidates being invited to next test phase

benjstephenson:


app/services/onlinetesting/Phase2TestService.scala
```diff
@@ -250,17 +250,24 @@ trait Phase2TestService extends OnlineTestService with ScheduleSelector {
...
-      require(u.phase2TestGroup.activeTests.nonEmpty, "Active tests cannot be found")
-      val activeTestsCompleted = u.phase2TestGroup.activeTests forall (_.completedDateTime.isDefined)
-      activeTestsCompleted match {
-        case true =>
-          phase2TestRepo.updateProgressStatus(u.applicationId, ProgressStatuses.PHASE2_TESTS_COMPLETED) map { _ =>
+  private def isTestGroupExpired(cubiksUserId: Int) = for {
+    profile <- phase2TestRepo.getTestProfileByCubiksId(cubiksUserId)
+  } yield profile.phase2TestGroup.expirationDate.isBeforeNow()
+
+  def markAsCompleted(cubiksUserId: Int)(implicit hc: HeaderCarrier, rh: RequestHeader): Future[Unit] = isTestGroupExpired(cubiksUserId).flatMap {
```
ollyw: It seems like doing the query to phase2TestRepo.getTestProfileByCubiksId(cubiksUserId) in the isTestGroupExpired method is introducing another call, and perhaps a race condition. Is it possible to put a query in the updatePhase2Test that only performs the update if necessary

app/repositories/onlinetesting/Phase3TestRepository.scala
```diff
@@ -57,7 +57,9 @@ class Phase3TestMongoRepository(dateTime: DateTimeFactory)(implicit mongo: () =>
   override implicit val bsonHandler: BSONHandler[BSONDocument, Phase3TestGroup] = Phase3TestGroup.bsonHandler
 
   override def nextApplicationsReadyForOnlineTesting: Future[List[OnlineTestApplication]] = {
-    val query = BSONDocument("applicationStatus" -> ApplicationStatus.PHASE2_TESTS_PASSED)
+    val query = BSONDocument("applicationStatus" -> ApplicationStatus.PHASE2_TESTS_PASSED,
+      s"progress-status.${PHASE2_TESTS_COMPLETED}" -> true
```
ollyw: How is the user getting in a situation that they are in ApplicationStatus.PHASE2_TESTS_PASSED without a progress status of PHASE2_TESTS_COMPLETED?

test/mocks/application/DocumentRootInMemoryRepository.scala
```diff
@@ -172,4 +173,6 @@ class DocumentRootInMemoryRepository extends GeneralApplicationRepository {
   def removeProgressStatuses(appId: String, progressStatuses: List[ProgressStatuses.ProgressStatus]): Future[Unit] = ???
 
   def getOnlineTestApplication(appId: String): Future[Option[OnlineTestApplication]] = ???
+
+  def progressCheck(applicationId: String, progressStatus: ProgressStatus): Future[Boolean] = ???
```
ollyw: If this is only used for test purposes, perhaps we could put it on the implementation with a protected access modifier?

app/services/onlinetesting/Phase2TestService.scala
benjstephenson: I thought about that but the updatePhase2Test method is called from a few places and uses several other functions to perform the update. I'll move it to the underlying repository method?

test/mocks/application/DocumentRootInMemoryRepository.scala
benjstephenson: I might just delete it - it was when I was thinking we needed to check the progress status when actually it's totally redundant in this case.

benjstephenson:
@ollyw comments addressed!  Hurrah!

app/repositories/onlinetesting/OnlineTestRepository.scala
```diff
@@ -95,10 +96,20 @@ trait OnlineTestRepository[U <: Test, T <: TestProfile[U]] extends RandomSelecti
...
 
   def updateTestCompletionTime(cubiksUserId: Int, completedTime: DateTime) = {
+    import repositories.BSONDateTimeHandler
+    val query = BSONDocument(s"testGroups.$phaseName.expirationDate" -> BSONDocument("$gt" -> DateTime.now(DateTimeZone.UTC)))
     val update = BSONDocument("$set" -> BSONDocument(
       s"testGroups.$phaseName.tests.$$.completedDateTime" -> Some(completedTime)
     ))
-    findAndUpdateCubiksTest(cubiksUserId, update)
+
+    play.api.Logger.debug(play.api.libs.json.Json.toJson(query).toString)
```
ollyw: This looks like it can be removed now

app/repositories/onlinetesting/Phase3TestRepository.scala
ollyw: Shouldn't the progress be be PHASE2_TESTS_PASSED? I think that is the terminal state for phase 2 where the test has passed

test/mocks/application/DocumentRootInMemoryRepository.scala
ollyw: Any further thoughts?

ollyw:
@benjstephenson I took another look at this. I am keen to get it merged this morning if possible. Can you have a look at the updated comments and see if they make sense

benjstephenson:
@ollyw removed the unused methods and fixed the progress statuses in the repo queries.

ollyw:
@benjstephenson there is a test failing in the phase2repository spec. If you fix that and sort the merge conflict, i would be happy to merge :-)

PR #151, opened 2016-10-27
FSET-932: Brief model changes for EDIP

satishkamatam:


app/model/ApplicationRoute.scala
```diff
@@ -0,0 +1,39 @@
...
+
+
+object ApplicationRoute extends Enumeration {
+
+  type ApplicationRoute = Value
+
+  val FASTSTREAM, EDIP, SDIP = Value
+
+  implicit val applicationRouteFormat = new Format[ApplicationRoute] {
+    def reads(json: JsValue) = JsSuccess(ApplicationRoute.withName(json.as[String].toUpperCase()))
```
benjstephenson: Why upper case? Other constants we have in mongo are lowercase (with the exception of the ProgressStatus fiasco).  Same for the value names.

PR #152, opened 2016-10-27
Use UTC in test to fix the issue with time change

lukastymo:


henricook:
You look good, get in here!

PR #154, opened 2016-10-28
Skip evaluation for unset passmarks (required by Edip/Sdip)

lukastymo:


app/services/onlinetesting/phase1/Phase1TestEvaluation.scala
```diff
@@ -18,23 +18,29 @@ package services.onlinetesting.phase1
...
-import model.exchange.passmarksettings.{ PassMarkThreshold, Phase1PassMarkSettings }
+import model.exchange.passmarksettings.{ PassMarkThreshold, Phase1PassMarkSettings, Phase1PassMarkThresholds }
 import model.persisted.{ SchemeEvaluationResult, TestResult }
 
 trait Phase1TestEvaluation {
 
   def evaluateForGis(schemes: List[SchemeType], sjqTestResult: TestResult, passmark: Phase1PassMarkSettings): List[SchemeEvaluationResult] = {
-    schemes map { scheme =>
-      val schemePassmark = passmark.findPassmarkForScheme(scheme)
+    for {
```
satishkamatam: shall we refactor this to extract into a method to be reused with gis and non gis

app/services/onlinetesting/phase1/Phase1TestEvaluation.scala
```diff
@@ -46,7 +52,7 @@ trait Phase1TestEvaluation {
         case (Green, Green) => Green
       }
 
-      SchemeEvaluationResult(scheme, result.toString)
+      SchemeEvaluationResult(schemeToEvaluate, result.toString)
```
satishkamatam: Though the change is not part of this PR ... any reason why we convert to result.toString, instead of using result which is type safe ??

lukastymo: I think this is inherited from fasttrack. Agree, It should be type safe

app/services/onlinetesting/phase1/Phase1TestEvaluation.scala
lukastymo: I was thinking about this. Now you assured me it's a good idea

PR #158, opened 2016-10-31
Create EDIP candidates with the test data generator

benjstephenson:


ollyw:
Putting a short term additional route is not ideal. Did you look at the possibility of using the POST route that @mgilgar has added, which removes the limitations that we have with mapping GETs onto case classes

benjstephenson:
I agree it's far from ideal.  I did initially but spent too much time fiddling with the various changes needed in the acceptance test client to get it to work.  I'll revisit this though one night soon.

PR #161, opened 2016-10-31
FSET-878: Add applicationRoute to Candidate to display the type in admin

lukastymo:


app/controllers/SearchForApplicantsController.scala
```diff
@@ -53,17 +54,21 @@ trait SearchForApplicantsController extends BaseController {
...
         }
       }.recover {
         case e: PersonalDetailsNotFound =>
-          Ok(Json.toJson(Candidate(userId, Some(application.applicationId), None, None, None, None, None, None, None, None)))
+          Ok(Json.toJson(Candidate(userId, Some(application.applicationId), None, None, None, None, None, None, None, None,
+            Some(application.applicationRoute))))
       }
     }.recover {
-      case e: ApplicationNotFound => Ok(Json.toJson(Candidate(userId, None, None, None, None, None, None, None, None, None)))
+      // when application is not found, the application route is set to Faststream for backward compatibility
```
ollyw: Is the plan to refactor this out at some point? It doesn't seem to make sense to hardcode to Faststream, given it is an Option anyway. Why not pass none?

app/repositories/application/GeneralApplicationRepoBSONToModelHelper.scala
```diff
@@ -107,23 +107,11 @@ trait GeneralApplicationRepoBSONToModelHelper {
     val applicationId = doc.getAs[String]("applicationId").getOrElse("")
     val progress: ProgressResponse = findProgress(doc, applicationId)
 
-    CandidateProgressReport(applicationId, Some(ProgressStatusesReportLabels.progressStatusNameInReports(progress)), schemes.getOrElse(List.empty[SchemeType]), disability, onlineAdjustments,
+    CandidateProgressReport(applicationId, Some(ProgressStatusesReportLabels.progressStatusNameInReports(progress)),
+      schemes.getOrElse(List.empty[SchemeType]), disability, onlineAdjustments,
```
ollyw: schemes.getOrElse(List.empty[SchemeType]) could be replaced with Nil here as the type can be inferred. I appreciate that you just reformatted the code, but in future we can use that for readability

app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -269,18 +269,25 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService,
 
   def findStatus(applicationId: String): Future[ApplicationStatusDetails] = {
     val query = BSONDocument("applicationId" -> applicationId)
-    val projection = BSONDocument("applicationStatus" -> 1, "progress-status-timestamp" -> 1, "progress-status-dates" -> 1, "_id" -> 0)
+    val projection = BSONDocument(
+      "applicationStatus" -> 1,
```
ollyw: I'll merge this one, but let's use true/false in future

PR #162, opened 2016-11-01
Extend a launchpad test

henricook:
An endpoint that allows a launchpad test to be extended

There was also some refactoring to audit and data store event creation/mocking in tests.


test/services/events/EventServiceSpec.scala
```diff
@@ -71,6 +71,12 @@ trait EventServiceFixture extends MockitoSugar {
     assert(eventCaptor.getAllValues.toList.forall(_.eventName == eventName))
   }
 
+  def verifyDataStoreEvents(n: Int, eventNames: List[String]): Unit = {
+    val eventCaptor = ArgumentCaptor.forClass(classOf[DataStoreEvent])
+    verify(dataStoreEventHandlerMock, times(n)).handle(eventCaptor.capture)(any[HeaderCarrier], any[RequestHeader])
+    assert(eventNames.forall(eventName => eventCaptor.getAllValues.toList.exists(_.eventName == eventName)))
```
lukastymo: - I think assert from Predef is not designed for unit testing. It gives you unclear error message because it just throw an exception on false.
- forall returns true for empty list, even though eventCaptor.getAllValues returns not empty list.

test/services/onlinetesting/Phase3TestServiceSpec.scala
```diff
@@ -131,7 +186,10 @@ class Phase3TestServiceSpec extends PlaySpec with MockitoSugar with ScalaFutures
 
     val testInterviewId = 123
     val testTimeNow = DateTime.parse("2016-10-01T00:00:01Z")
-    val testExpiryTime = testTimeNow.plusDays(7)
+    val unexpiredTestExpiryTime = DateTime.parse("2016-11-01T00:00:01Z")
```
lukastymo: Can't we mock dateTimeFactory and return always date which we know (localdate.now()). It would make the tests not bound to this exact date.

If it will be a lot of work leave it as it is.

app/services/onlinetesting/Phase3TestService.scala
```diff
@@ -134,11 +142,39 @@ trait Phase3TestService extends OnlineTestService with Phase3TestConcern with Re
...
+    val phase3TestGroup = phase3TestRepo.getTestGroup(applicationId)
+      .map(tg => tg.getOrElse(throw new IllegalStateException("Expiration date for Phase 3 cannot be extended. Test group not found.")))
+
+    for {
+      progress <- progressFut
+      phase3 <- phase3TestGroup
+      isAlreadyExpired = progress.phase3ProgressResponse.phase3TestsExpired
+      extendDays = extendTime(isAlreadyExpired, phase3.expirationDate)
+      newExpiryDate = extendDays(extraDays)
+      activeTest = phase3.activeTests.head
```
henricook: Created .activeTest which has a require() in

PR #166, opened 2016-11-02
Bug FSET-933: e-Tray link missing on candidates dashboard

Lansalo:


app/services/application/ApplicationService.scala
```diff
@@ -110,4 +110,41 @@ trait ApplicationService extends EventSink {
     }.map(_ => ())
   }
 
+
+  def fix(toBeFixed: Seq[FixRequiredType])(implicit hc: HeaderCarrier, rh: RequestHeader): Future[Unit] = {
+    Future { toBeFixed.foreach(fixData(_)) }
```
ollyw: We definitely should use `FutureEx.traverseSerial` here, rather than orphan futures with `foreach`

PR #167, opened 2016-11-03
Update SBT version to latest

henricook:


ollyw:
@henricook exciting times. Next stop SBT 1.0

PR #169, opened 2016-11-03
FSET 691: Invite candidates to take the video interview (minor tweaks)

henricook:
RTM

Riskiest part of the PR: Refactors the e-tray (phase2) invite logic
.. the rest is fairly innocuous for the moment


app/repositories/CommonBSONDocuments.scala
```diff
@@ -60,6 +60,37 @@ trait CommonBSONDocuments {
     )
   }
 
+  def inviteToTestBSON[P <: ProgressStatus](targetProgressStatus: P, invigilatedKeyToExclude: String) = {
```
lukastymo: it's not entirely common. do we need a new common trait just for tests? Otherwise we will end up having something like: OnlineTestApplication (which sounds like common thing, but contains things connected with etray, video interview etc.)

PR #170, opened 2016-11-03
Fix multiple submission

ollyw:
- Guard against multiple application submissions
- Ensure the application is in the right state to be submitted


henricook:
@ollyw Seems to have a recurring test failure

benjstephenson:
Tests all pass locally, LGTM 👍

ollyw:
Ok, so there have been some changes which should fix these tests

PR #180, opened 2016-11-07
FSET-719 set etray passmarks

benjstephenson:


app/controllers/Phase2PassMarkSettingsController.scala
```diff
@@ -0,0 +1,72 @@
...
+  val uuidFactory: UUIDFactory
+
+  def create = Action.async(parse.json) { implicit request =>
+    withJsonBody[Phase2PassMarkSettings] { passMarkSettings => {
+        val newVersionUUID = uuidFactory.generateUUID()
+        val newPassMarkSettings = passMarkSettings.copy(version = newVersionUUID, createDate = DateTime.now())
+        for {
+          createResult <- passMarkService.createPassMarkSettings(newPassMarkSettings)
+        } yield {
+          auditService.logEvent("PassMarkSettingsCreated", Map(
```
satishkamatam: Do we need to be specific in creating the event PassMarkSettingsCreated -> Phase1 or Phase2

app/model/exchange/passmarksettings/PassMarkTestTypes.scala
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package model.exchange.passmarksettings
+
+import play.api.libs.json.{ Format, JsString, JsSuccess, JsValue }
+import reactivemongo.bson.{ BSON, BSONHandler, BSONString }
+
+object PassMarkTestTypes extends Enumeration {
```
satishkamatam: Shall we delete this if not used

app/controllers/Phase2PassMarkSettingsController.scala
benjstephenson: We probably should - I'll update the phase 1 controller too.

app/model/exchange/passmarksettings/PassMarkTestTypes.scala
benjstephenson: Yes

PR #185, opened 2016-11-09
Fset 691: Further work, make invites respect time adjustments

henricook:


ollyw:
Awesome, you are knocking it out the park! LGTM

PR #199, opened 2016-11-11
FSET 869: Store 'reviewed' callbacks from Launchpad

henricook:


app/connectors/launchpadgateway/exchangeobjects/in/reviewed/ReviewedCallbackRequest.scala
```diff
@@ -0,0 +1,62 @@
...
+    scoreForQuestion(1, latestReviewer.question1) +
+    scoreForQuestion(2, latestReviewer.question2) +
+    scoreForQuestion(3, latestReviewer.question3) +
+    scoreForQuestion(4, latestReviewer.question4) +
+    scoreForQuestion(5, latestReviewer.question5) +
+    scoreForQuestion(6, latestReviewer.question6) +
+    scoreForQuestion(7, latestReviewer.question7) +
+    scoreForQuestion(8, latestReviewer.question8)
+  }
+}
```
henricook: @ollyw I think this is good enough, it will let us fetch a test (with its callbacks) and do dynamic scoring.

I've abandoned the scoring-as-callbacks-come-in idea because in some situations i don't think we want to calculate a score (see LaunchpadQuestionIsUnscoredException above). I think that when the score processor comes to a user it should call this method and if it gets the exception carry on to come back to this user later.

The reviewer team will proactively contact the candidate via Launchpad support who will prompt them for resets/to retake questions that cannot be scored, the assumption is that ultimately no one will have an unscored question - and if they do we won't process them. Need to check this with Mike, there's a ticket on the board about it.

PR #201, opened 2016-11-14
Suppress email invites for invigilated candidates.

benjstephenson:


henricook:
If tests pass you're in!

PR #203, opened 2016-11-14
FSET-814: access code functionality for invigilated e-tray

ian-stainer:
Added the verify access code functionality for invigilated e-tray


app/repositories/contactdetails/ContactDetailsRepository.scala
```diff
@@ -60,4 +62,15 @@ class ContactDetailsMongoRepository(implicit mongo: () => DB)
       case None => throw ContactDetailsNotFound(userId)
     }
   }
+
+  def findUserIdByEmail(email: String): Future[String] = {
+    val query = BSONDocument("contact-details.email" -> email)
```
benjstephenson: Do we need to ensure case sensitivity here?  That may be taken care of by the FE caller.

app/repositories/onlinetesting/Phase2TestRepository.scala
```diff
@@ -39,6 +39,8 @@ trait Phase2TestRepository extends OnlineTestRepository with Phase2TestConcern {
 
   def getTestGroup(applicationId: String): Future[Option[Phase2TestGroup]]
 
+  def getTestGroupByUserId(userId: String): Future[Option[Phase2TestGroup]]
```
benjstephenson: We could probably move this and the implementation up to the OnlineTestRepository as it is generic enough.

app/services/onlinetesting/Phase2TestService.scala
```diff
@@ -95,6 +95,25 @@ trait Phase2TestService extends OnlineTestService with Phase2TestConcern with Sc
     }
   }
 
+  def verifyAccessCode(email: String, accessCode: String): Future[Option[String]] = {
+    (for {
+      userId <- cdRepository.findUserIdByEmail(email)
+      testGroupOpt <- phase2TestRepo.getTestGroupByUserId(userId)
+    } yield {
+      testGroupOpt.flatMap { testGroup =>
+        val eTrayTest = testGroup.activeTests.head
```
benjstephenson: I know that technically this will always be a list of 1 item for etray - but what if it's empty 😱

app/services/onlinetesting/Phase2TestService.scala
```diff
@@ -95,6 +95,25 @@ trait Phase2TestService extends OnlineTestService with Phase2TestConcern with Sc
     }
   }
 
+  def verifyAccessCode(email: String, accessCode: String): Future[Option[String]] = {
+    (for {
+      userId <- cdRepository.findUserIdByEmail(email)
+      testGroupOpt <- phase2TestRepo.getTestGroupByUserId(userId)
+    } yield {
+      testGroupOpt.flatMap { testGroup =>
```
benjstephenson: We could bump this up to the comprehension to remove one level of nesting and explicit mapping?

PR #207, opened 2016-11-15
fset-814 added expired test scenario

Lansalo:
Added logic to take care of an expired phase 2 test for invigilated e-tray.
In this case a 403 will be returned back to the client, rather than the 404 for failed verification.

### DIFF
--- a/app/services/onlinetesting/Phase2TestService.scala
+++ b/app/services/onlinetesting/Phase2TestService.scala
@@ -96,14 +96,11 @@ trait Phase2TestService extends OnlineTestService with Phase2TestConcern with Sc
     }
   }
 
-  def verifyAccessCode(email: String, accessCode: String): Future[Try[String]] = {
-    (for {
-      userId <- cdRepository.findUserIdByEmail(email)
-      testGroupOpt <- phase2TestRepo.getTestGroupByUserId(userId)
-    } yield processEtrayToken(testGroupOpt, accessCode)) recover {
-      case e: ContactDetailsNotFoundForEmail => Failure(e)
-    }
-  }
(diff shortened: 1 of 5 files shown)


app/services/onlinetesting/Phase2TestService.scala
```diff
@@ -95,22 +96,12 @@ trait Phase2TestService extends OnlineTestService with Phase2TestConcern with Sc
     }
   }
 
-  def verifyAccessCode(email: String, accessCode: String): Future[Option[String]] = {
+  def verifyAccessCode(email: String, accessCode: String): Future[Try[String]] = {
```
benjstephenson: Why Try and not Future[String]?

Lansalo: Just because I found it easier to refactor the existing code with a Try[String].
Not to say that it couldn't have been done with Future[String]

ollyw: I think @benjstephenson has a point. Future has the concept of Failure baked in, and putting a Try in it just makes the caller work harder. If this were a private method it would not be such an issue, but muddying a public API because of the internal implementation is really a last resort. Perhaps using Future.fromTry would enable you to use your for comprehension over futures to simplify. Something like:
```
  def verifyAccessCode(email: String, accessCode: String): Future[String] = {
    for {
      userId <- cdRepository.findUserIdByEmail(email)
      testGroupOpt <- phase2TestRepo.getTestGroupByUserId(userId)
      value <- Future.fromTry(processEtrayToken(testGroupOpt, accessCode))
    } yield value
  }
```

PR #213, opened 2016-11-16
Preferred names in the TDG were being randomised, which was confusing. They are now based on first names

henricook:


ollyw:
LTGM 👍

PR #216, opened 2016-11-17
FSET-693: Notify candidate of a video interview failed result

Lansalo:


ollyw:
Perfect. @Lansalo you are getting quick at these notification jobs

PR #219, opened 2016-11-18
FSET-989: Invigilated video invites (And Mockito 2)

henricook:


test/testkit/MockitoSugar.scala
```diff
@@ -23,8 +23,8 @@ import scala.concurrent.Future
 
 trait MockitoSugar extends org.scalatest.mock.MockitoSugar {
   // Unfortunately we cannot use the `when().thenReturn` format for spies.
-  def doReturnAsync() = Mockito.doReturn(Future.successful(()))
-  def doReturnAsync[A](value: A) = Mockito.doReturn(Future.successful(value))
+  def doReturnAsync() = Mockito.doReturn(Future.successful(()), Nil.toSeq:_*)
```
henricook: This is the only Mockito 2 strangeness i encountered. doReturn now has a second, varArgs* argument, and the compiler had conflicts when only one argument was specified. I've specified an empty varArgs set as a result, but i'd be open to anyone who knows a more concise way of doing that than Nil.toSeq:_*

app/services/onlinetesting/Phase3TestService.scala
```diff
@@ -44,14 +43,13 @@ import services.onlinetesting.Phase2TestService.NoActiveTestException
 import uk.gov.hmrc.play.http.HeaderCarrier
 
 import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Future
+import scala.concurrent.{ Await, Future }
```
ollyw: Seek and destroy uses of `Await`

app/services/onlinetesting/Phase3TestService.scala
```diff
@@ -251,7 +263,8 @@ trait Phase3TestService extends OnlineTestService with Phase3TestConcern {
                                (implicit hc: HeaderCarrier, rh: RequestHeader): Future[Unit] = {
     val progressFut = appRepository.findProgress(applicationId)
     val phase3TestGroup = phase3TestRepo.getTestGroup(applicationId)
-      .map(tg => tg.getOrElse(throw new IllegalStateException("Expiration date for Phase 3 cannot be extended. Test group not found.")))
+      .map(tg => tg.getOrElse(throw new IllegalStateException(s"Phase 3 tests Expiration date for app id: '$applicationId' " +
+        s"cannot be extended. Test group not found.")))
```
ollyw: you can save an `s` here FWIW

app/services/onlinetesting/Phase3TestService.scala
```diff
@@ -66,7 +64,6 @@ object Phase3TestService extends Phase3TestService {
 }
 
 trait Phase3TestService extends OnlineTestService with Phase3TestConcern {
-  val adjustmentsService: AdjustmentsManagementService
```
ollyw: Seems like there are quite a few dependencies mixed in from OnlineTestService that don't need to be redeclared here
```
  val phase3TestRepo: Phase3TestRepository
  val launchpadGatewayClient: LaunchpadGatewayClient
  val tokenFactory: UUIDFactory
  val dateTimeFactory: DateTimeFactory
  val emailClient: OnlineTestEmailClient
  val auditService: AuditService
  val gatewayConfig: LaunchpadGatewayConfig
```
could go to
```
  val phase3TestRepo: Phase3TestRepository
  val launchpadGatewayClient: LaunchpadGatewayClient
  val gatewayConfig: LaunchpadGatewayConfig
```

app/services/onlinetesting/Phase3TestService.scala
```diff
@@ -118,19 +115,34 @@ trait Phase3TestService extends OnlineTestService with Phase3TestConcern {
...
+      case true => extendTestGroupExpiryTime(application.applicationId, daysUntilExpiry - 7, "InvigilatedInviteSystem")
+      case false => Future.successful(())
+    }
+
     for {
       emailAddress <- candidateEmailAddress(application)
       phase3Test <- registerAndInviteApplicant(application, emailAddress, interviewId, invitationDate, expirationDate)
       _ <- emailProcess(emailAddress)
       _ <- markAsInvited(application)(Phase3TestGroup(expirationDate = expirationDate, tests = List(phase3Test)))
+      _ <- extendIfInvigilated(application)
```
ollyw: Should we/can we extend before marking as invited. If `extendIfInvigilated` fails due to mongo or other things, it would seem like the candidate is good to go

test/testkit/MockitoSugar.scala
ollyw: That is rather weird! First I considered that .toSeq seems redundant, and can be removed. Then I dug a bit deeper and considered it a compiler bug until discovered the discussion here http://stackoverflow.com/questions/13358705/force-single-argument-in-scala-varargs/14290755#14290755

app/services/onlinetesting/Phase3TestService.scala
henricook: Such win

https://goo.gl/Sf2nX

PR #223, opened 2016-11-18
FSET-723: Video interview pass mark settings and refactoring

satishkamatam:


app/controllers/PassMarkSettingsController.scala
```diff
@@ -18,36 +18,61 @@ package controllers
...
 import scala.concurrent.ExecutionContext.Implicits.global
 
-object Phase1PassMarkSettingsController extends Phase1PassMarkSettingsController {
+object Phase1PassMarkSettingsController extends PassMarkSettingsController[Phase1PassMarkSettings] {
   val passMarkService = Phase1PassMarkSettingsService
   val auditService = AuditService
   val uuidFactory = UUIDFactory
+  val passMarksCreatedEvent = "Phase1PassMarkSettingsCreated"
+  def getCopy(passMarkSettings:Phase1PassMarkSettings, newVersionUUID: String) =
+    passMarkSettings.copy(version = newVersionUUID, createDate = DateTime.now())
```
mgilgar: I guess it is not a copy if you change the version, does it sound better to say "upgradeVersion" (if that is the intention)?

app/controllers/PassMarkSettingsController.scala
```diff
@@ -18,36 +18,61 @@ package controllers
...
 }
 
-trait Phase1PassMarkSettingsController extends BaseController {
+object Phase2PassMarkSettingsController extends PassMarkSettingsController[Phase2PassMarkSettings] {
+  val passMarkService = Phase2PassMarkSettingsService
+  val auditService = AuditService
+  val uuidFactory = UUIDFactory
+  val passMarksCreatedEvent = "Phase2PassMarkSettingsCreated"
+  def getCopy(passMarkSettings:Phase2PassMarkSettings, newVersionUUID: String) =
```
mgilgar: Some comment
[I guess it is not a copy if you change the version, does it sound better to say "upgradeVersion" (if that is the intention)?]

app/controllers/PassMarkSettingsController.scala
```diff
@@ -18,36 +18,61 @@ package controllers
...
+    passMarkSettings.copy(version = newVersionUUID, createDate = DateTime.now())
+}
 
-  val passMarkService: PassMarkSettingsService[Phase1PassMarkSettings]
+object Phase3PassMarkSettingsController extends PassMarkSettingsController[Phase3PassMarkSettings] {
+  val passMarkService = Phase3PassMarkSettingsService
+  val auditService = AuditService
+  val uuidFactory = UUIDFactory
+  val passMarksCreatedEvent = "Phase3PassMarkSettingsCreated"
+  def getCopy(passMarkSettings:Phase3PassMarkSettings, newVersionUUID: String) =
```
mgilgar: Some comment
[I guess it is not a copy if you change the version, does it sound better to say "upgradeVersion" (if that is the intention)?]

app/controllers/PassMarkSettingsController.scala
```diff
@@ -18,36 +18,61 @@ package controllers
...
+
+abstract class PassMarkSettingsController[T <: PassMarkSettings]
+(implicit manifest: Manifest[T], jsonFormat: Format[T]) extends BaseController {
+
+  val passMarkService: PassMarkSettingsService[T]
   val auditService: AuditService
   val uuidFactory: UUIDFactory
+  val passMarksCreatedEvent : String
+
+  def getCopy(passMarkSettings:T, newVersionUUID: String) : T
```
mgilgar: Some comment
[I guess it is not a copy if you change the version, does it sound better to say "upgradeVersion" (if that is the intention)?]

app/model/exchange/passmarksettings/PassMarkSettings.scala
```diff
@@ -52,3 +52,16 @@ object Phase2PassMarkSettings {
   implicit val phase2PassMarkSettingsFormat = Json.format[Phase2PassMarkSettings]
   implicit val phase2PassMarkSettingsHandler = Macros.handler[Phase2PassMarkSettings]
 }
+
+case class Phase3PassMarkSettings(
+  schemes: List[Phase3PassMark],
```
mgilgar: The way I see Phase3PassMarkSettings is that it is a list of passmarks (which is schemes field) and then some metadata. So schemes field is actually the passmarkSettings without metadata. In that sense calling them schemes is a bit misleading, I would exactly what it is exactly is passmarkSettingsPerScheme or passmarksPerScheme

app/model/exchange/passmarksettings/PassMarkThresholds.scala
```diff
@@ -45,6 +49,15 @@ case class Phase2PassMarkThresholds(
 ) extends PassMarkThresholds with Phase2Thresholds
 
 object Phase2PassMarkThresholds {
-  implicit val phase1PassMarkThresholds = Json.format[Phase2PassMarkThresholds]
-  implicit val phase1PassMarkThresholdsHandler = Macros.handler[Phase2PassMarkThresholds]
+  implicit val phase2PassMarkThresholds = Json.format[Phase2PassMarkThresholds]
+  implicit val phase2PassMarkThresholdsHandler = Macros.handler[Phase2PassMarkThresholds]
+}
+
+case class Phase3PassMarkThresholds(
```
mgilgar: It feels redundant to wrap videoInterview in another case class when there is only one field but I guess you want to keep consistency with the other Phase2 and Phase1...

test/controllers/Phase1PassMarkSettingsControllerSpec.scala
```diff
@@ -39,7 +39,8 @@ class Phase1PassMarkSettingsControllerSpec extends UnitWithAppSpec {
     "Return a settings objects with schemes but no thresholds if there are no settings saved" in new TestFixture {
```
mgilgar: Does the test corresponds to the description? Where do we assert that we get "a settings objects with schemes but no thresholds"?

app/model/exchange/passmarksettings/PassMarkThresholds.scala
satishkamatam: yes for consistency reasons

test/controllers/Phase1PassMarkSettingsControllerSpec.scala
satishkamatam: Looks like the test description was wrong. changed.

app/model/exchange/passmarksettings/PassMarkSettings.scala
satishkamatam: schemes is a bit misleading agreed, but given that this is a field associated with Phase3PassMarkSettings there shouldn't be too much of confusion. Also i want to keep it the same way as other phases

app/controllers/PassMarkSettingsController.scala
satishkamatam: renamed to upgradeVersion

app/controllers/PassMarkSettingsController.scala
satishkamatam: upgradeVersion

app/controllers/PassMarkSettingsController.scala
satishkamatam: renamed to upgradeVersion

app/model/exchange/passmarksettings/PassMarkSettings.scala
mgilgar: Do we use schemes field name in the other phases?

PR #224, opened 2016-11-18
FSET-944: Send email when adjustments has been updated

lukastymo:


test/services/events/EventServiceSpec.scala
```diff
@@ -44,7 +44,7 @@ class EventServiceSpec extends UnitSpec with EventServiceFixture {
   }
 }
 
-trait EventServiceFixture extends MockitoSugar {
+trait EventServiceFixture extends MockitoSugar with org.scalatest.MustMatchers {
```
henricook: Why the fully qualified package name instead of an import?

PR #228, opened 2016-11-23
FSET-792: Calculate and display the North/South indicator

Lansalo:
This PR is related to the corresponding change for the admin frontend: https://github.tools.tax.service.gov.uk/HMRC/fset-faststream-admin-frontend/pull/105


app/controllers/ReportingController.scala
```diff
@@ -120,6 +130,13 @@ trait ReportingController extends BaseController {
     }
   }
 
+  private def enrichReport(candidates: List[CandidateProgressReportItem], postcodes: Map[String, String]):
+  Future[List[CandidateProgressReportItem]] = {
+
+    Future(candidates.map(candidate => candidate.copy(
```
Lansalo: My intention was to actually spin this in another thread and it might be quite demanding from a computational point of view. The postcode map where we do the lookup, will have basically all the users with a postcode so could be quite chubby. But if you think that might not be an issue, I'm happy to use Future.successful

app/repositories/NorthSouthIndicatorCSVRepository.scala
```diff
@@ -0,0 +1,74 @@
...
+
+      values.foldLeft(Map.empty[String, FSACIndicator])((acc, line) => toMap(acc, line))
+    }
+  }
+
+  override def calculateFsacIndicator(postcode: Option[String], outsideUk: Boolean): Option[String] = {
+    (postcode, outsideUk) match {
+      case (None, false) => None
+      case (None, true) => Some(DefaultIndicator)
+      case (Some(pc), _) => {
```
Lansalo: That should never happen as during registration we allow user to enter the postcode if and only if inside the UK. But I can add false to make it more robust.

app/controllers/ReportingController.scala
ollyw: The thread is already a pool thread, and it is not done in parallel either, so it doesn't seem to be provide any benefit.  Have you benchmarked this for performance? If there is a performance issue, it may be better to pass the location lookup map into `reportRepository.candidateProgressReport` so that no copying is required

test/controllers/report/ReportingControllerSpec.scala
```diff
@@ -102,225 +112,115 @@ class ReportingControllerSpec extends UnitWithAppSpec {
     }
   }
 
-/*
```
Lansalo: Brought the commented out code back

app/repositories/NorthSouthIndicatorCSVRepository.scala
Lansalo: Was wrong I'll have to leave it like:
case (Some(pc), _) =>
That's because the outside/inside UK flag is not returned as part of candidateProgressReport.
I might modify findAllPostcodes to return that flag as well but it seemed over-engineered as a solution as, as I mentioned, we prevent candidate from outside UK to enter a postcode as part of the registration.

test/controllers/report/ReportingControllerSpec.scala
ollyw: Cool, thanks

test/controllers/report/ReportingControllerSpec.scala
```diff
@@ -310,11 +373,60 @@ class ReportingControllerSpec extends UnitWithAppSpec {
...
+          Some("No"), Some("No"), Some("No"), Some("No"), Some("No"), Some("No"), Some("No"), Some("No"), Some("1234567"), None, Some("Edip")),
+        CandidateProgressReportItem("user4", "app4", Some("submitted"),
+          List(SchemeType.DiplomaticService, SchemeType.GovernmentOperationalResearchService), Some("Yes"),
+          Some("No"), Some("No"), Some("No"), Some("No"), Some("No"), Some("No"), Some("No"), Some("No"), Some("1234567"), None, None)
+      )
+    )
+
+    val SuccessfulFindAllPostCodeResponse = Future.successful(
+      Map(
+        "user1" -> "EH2 4AD",
```
ollyw: We shouldn't be putting real data in the unit tests. I appreciate we need a matching prefix though. Postcodes such as EH9 9ZZ don't exist yet (and probably never will)

app/repositories/NorthSouthIndicatorCSVRepository.scala
ollyw: if we are not actually using the boolean, perhaps it shouldn't even be in the method signature, otherwise it is a bit misleading. If we are actually passing the outsideUK in at some call sites, but not others, that is a bit risky

PR #229, opened 2016-11-23
FSET-726: Expire Phase 3 Jobs

henricook:


test/scheduler/onlinetesting/ExpireOnlineTestJobSpec.scala
```diff
@@ -45,7 +45,7 @@ class ExpireOnlineTestJobSpec extends BaseServiceSpec with ShortTimeout {
     def interval: FiniteDuration = mock[FiniteDuration]
   }
 
-  "expire test phase 1 job" should {
+  "expire test phase N job" should {
```
ollyw: Perhaps this still WIP? The actual test still only works with Phase1, as according to TestableExpireTestJob, and the Phase2ExpirationEvent, Phase3ExpirationEvent imports are not being used

ollyw:
This all looks good. One minor query about around unit testing

test/scheduler/onlinetesting/ExpireOnlineTestJobSpec.scala
henricook: I can tidy the exports up, but when i put the tests in for those two events they didn't seem worth it, they weren't really testing anything. What do you think?

PR #233, opened 2016-11-25
Handle the update calls to mongo in a more robust way

ollyw:
This improves the way that mongo updates are handled. It is a WIP as there is still an outstanding question of what happens when we try to update, but the values are the same, and it appears that the nModified value doesn't reflect that.
A few tests were actually not working, and this change picked that up.
Also, the upsert flag was set in places that it should not be, especially went dealing with the application collection


app/repositories/ReactiveRepositoryHelpers.scala
```diff
@@ -0,0 +1,85 @@
...
+    singleUpdateValidatorImpl(id, actionDesc, ignoreNotFound = true, new Exception, upsert = true) _
+  }
+
+  def singleRemovalValidator(id: String, actionDesc: String): WriteResult => Unit = (result: WriteResult) => {
+    if (result.ok) {
+      if (result.n == 1) {
+        ()
+      } else if (result.n == 0) {
+        throw new NotFoundException(s"No document found whilst $actionDesc for id $id")
+      } else if (result.n > 1) {
```
benjstephenson: should we look at using the firstMatchOnly flag on collection.remove call to try and avoid deleting too many documents?

app/repositories/ReactiveRepositoryHelpers.scala
```diff
@@ -0,0 +1,85 @@
...
+
+trait ReactiveRepositoryHelpers {
+  this: ReactiveRepository[_, _] =>
+
+  protected lazy val bsonCollection = collection.db.collection[BSONCollection](collection.name)
+
+  def singleUpdateValidator(id: String,
+                            actionDesc: String,
+                            notFound: => Exception): UpdateWriteResult => Unit = {
+    singleUpdateValidatorImpl(id, actionDesc, ignoreNotFound = false, notFound, upsert = false) _
```
benjstephenson: Do we need these place holders at the end?

it/repositories/ReactiveRepositoryHelpersSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+import model.Exceptions.{ NotFoundException, NotModifiedException }
+import play.api.libs.json.Json
+import reactivemongo.bson._
+import reactivemongo.api.DB
+import reactivemongo.bson.{ BSONDocument, BSONObjectID }
+import testkit.MongoRepositorySpec
+import uk.gov.hmrc.mongo.ReactiveRepository
+import uk.gov.hmrc.mongo.json.ReactiveMongoFormats
+
+class ReactiveRepositoryHelpersSpec extends MongoRepositorySpec {
```
benjstephenson: No tests for singleRemovalValidator?

benjstephenson:
@ollyw minor comments, but otherwise LGTM and I'll merge.

it/repositories/ReactiveRepositoryHelpersSpec.scala
ollyw: The singleRemovalValidator is only being used in dormant code, but I have now added the test for the future

PR #235, opened 2016-11-28
Fset 931 - Reset etray

kjharkin:


app/repositories/onlinetesting/Phase3TestRepository.scala
```diff
@@ -120,6 +122,31 @@ class Phase3TestMongoRepository(dateTime: DateTimeFactory)(implicit mongo: () =>
     }
   }
 
+  def removeTestGroup(applicationId: String): Future[Unit] = {
+
+    val phase3Progresses = ProgressStatuses.allStatuses.filter(_.startsWith(phaseName))
```
ollyw: Selecting progress statuses by name is very brittle. If we have to have logic like this, perhaps it could be in the ProgressStatuses type, so at least people see it when they are changing progress statuses. Also, each progress status has an applicationStatus val, so we could filter without losing type safety

PR #236, opened 2016-11-28
Test data generator phase3 tests results received

mgilgar:
Changes to created phase3 callbacks in mongo for test data generator.
This is not part of any story but it will be useful in fset-724


ollyw:
LGTM 👍

PR #241, opened 2016-11-30
Fset 980: Notify candidate of video interview success

Lansalo:
This PR implements the notification for a successful video interview. It includes also some refactoring that should move us toward a unified idea of a candidate notification for some event (although some more refactoring will be needed for that).
It also add some extra unit coverage and fir an issue with wrong event reporting.

Note: related to https://github.tools.tax.service.gov.uk/HMRC/fset-faststream-admin-frontend/pull/112


test/mocks/application/DocumentRootInMemoryRepository.scala
```diff
@@ -113,8 +113,8 @@ class DocumentRootInMemoryRepository extends GeneralApplicationRepository {
 
   override def rejectAdjustment(applicationId: String): Future[Unit] = Future.successful(Unit)
 
-  override def findFailedTestForNotification(failedTestType: FailedTestType): Future[Option[NotificationFailedTest]] = {
-    Future.successful(Some(NotificationFailedTest("31009ccc-1ac3-4d55-9c53-1908a13dc5e1", "fbb466a3-13a3-4dd0-93d6-9dfa764a5555", "George")))
+  override def findTestForNotification(notificatioType: NotificationTestType): Future[Option[TestResultNotification]] = {
```
benjstephenson: notificationType?

app/services/onlinetesting/OnlineTestService.scala
```diff
@@ -61,25 +62,20 @@ trait OnlineTestService extends TimeExtension with EventSink {
...
   }
 
-  protected def processFailedTest(toNotify: NotificationFailedTest, failedType: FailedTestType)
-                                  (implicit hc: HeaderCarrier, rh: RequestHeader): Future[Unit] = {
-    def event(email: String): EventType = AuditEvents.FailedTestEmailSent(Map("applicationId" -> toNotify.applicationId,
-      "emailAddress" -> email,
-      "to" -> toNotify.preferredName,
-      "template" -> failedType.template))
-
+  protected def processTestForNotification(toNotify: TestResultNotification, notiType: NotificationTestType)
```
benjstephenson: I think we should prefer full words! Can we rename notiType?

app/model/NotificationTestType.scala
```diff
@@ -17,11 +17,20 @@
 package model
 
 import model.ApplicationStatus.ApplicationStatus
-import model.ApplicationStatus.{ PHASE1_TESTS_FAILED, PHASE2_TESTS_FAILED, PHASE3_TESTS_FAILED}
+import model.ApplicationStatus.{ PHASE1_TESTS_FAILED, PHASE2_TESTS_FAILED, PHASE3_TESTS_FAILED, READY_FOR_EXPORT }
 import model.ProgressStatuses._
 
+sealed trait NotificationTestType {
```
benjstephenson: I think this trait name is a bit confusing - its a notification for a test, not a test notification.  Maybe NotificationForTest?  Not sure the `Type` on end adds anything either.

app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -364,15 +365,30 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService,
...
     )))
 
-  override def findFailedTestForNotification(failedTestType: FailedTestType): Future[Option[NotificationFailedTest]] = {
-    val query = BSONDocument("$and" -> BSONArray(
-      BSONDocument("applicationStatus" -> failedTestType.appStatus),
-      BSONDocument(s"progress-status.${failedTestType.notificationProgress}" -> BSONDocument("$ne" -> true)),
-      BSONDocument(s"progress-status.${failedTestType.receiveStatus}" -> true)
-    ))
+  override def findTestForNotification(notificationType: NotificationTestType): Future[Option[TestResultNotification]] = {
+    val query = Try{ notificationType match {
```
benjstephenson: Do we need this Try block?  There are only success or fail types defined so we should never fall into the unknown match.  I think we'll get an inexhaustive match compile error if this isn't updated if we add more types anyway.

benjstephenson:
@Lansalo few comments

app/model/NotificationTestType.scala
Lansalo: The idea is that you can have different types of notification (failed, success, expired). Let me know if it make sense otherwise I'll change it

app/repositories/application/GeneralApplicationRepository.scala
Lansalo: At the present time we have just two types so you'll never end up in the unknown case, but this check is from when (and if) we'll add more of them. In that scenario you won't get any compile error but just a warning that the match is not exhaustive and then a runtime exception because of the unmatched case. Try is there just to wrap that exception in a Future failed rather than propagate it. Let me know if that is ok with you.

benjstephenson: I'm happy to l.eave it in if we think we're going to add more notification types then

PR #242, opened 2016-11-30
Fset 896 test data generator improvement

mgilgar:
These PR should be done together:
https://github.com/hmrc/fset-faststream/pull/238
https://github.com/hmrc/fset-faststream/pull/242
and 58 PR in candidate-acceptance-test
and 45 PR in admin-acceptance-test

### DIFF
--- a/app/model/Exceptions.scala
+++ b/app/model/Exceptions.scala
@@ -72,7 +72,7 @@ object Exceptions {
 
   case class IncorrectStatusInApplicationException(reason: String) extends Exception(reason)
 
-  case class InvalidStatusAndProgressStatusException(reason: String) extends Exception(reason)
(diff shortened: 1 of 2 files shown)


app/model/Exceptions.scala
```diff
@@ -72,7 +72,7 @@ object Exceptions {
 
   case class IncorrectStatusInApplicationException(reason: String) extends Exception(reason)
 
-  case class InvalidStatusException(reason: String) extends Exception(reason)
+  case class InvalidStatusAndProgressStatusException(reason: String) extends Exception(reason)
```
mgilgar: What I am trying to say with that exception is review your applicationStatus parameter and you progressStatus param because you pass a wrong combination.

mgilgar: Do you mean it should be InvalidApplicationStatusAndProgressStatusException?

PR #249, opened 2016-12-02
FSET-880: Add Faststream becomes SDIP path

lukastymo:


app/connectors/EmailClient.scala
```diff
@@ -198,6 +198,9 @@ trait EmailClient extends WSHttp {
     sendEmail(to, "fset_faststream_adjustments_changed",
       Map("name" -> name, "etrayAdjustments" -> etrayAdjustments, "videoAdjustments" -> videoAdjustments))
 
+  def sendApplicationExtendedToSdip(to: String, name: String)(implicit hc: HeaderCarrier) =
```
mgilgar: I would suggest to use the same method name as the email template, i.e.:

def sendApplicationConvertedToSdip(to: String, name: String)(implicit hc: HeaderCarrier) =

app/model/events/EmailEvents.scala
```diff
@@ -31,4 +31,5 @@ object EmailEvents {
   case class ApplicationWithdrawn(to: String, name: String) extends EmailEvent
   case class AdjustmentsConfirmed(to: String, name: String, etrayAdjustments: String, videoAdjustments: String) extends EmailEvent
   case class AdjustmentsChanged(to: String, name: String, etrayAdjustments: String, videoAdjustments: String) extends EmailEvent
+  case class ApplicationExtendedToSdip(to: String, name: String) extends EmailEvent
```
mgilgar: I think we have to choose here if we want to use converted or extended, or is it different?

it/repositories/application/GeneralApplicationMongoRepositorySpec.scala
```diff
@@ -469,6 +469,41 @@ class GeneralApplicationMongoRepositorySpec extends MongoRepositorySpec with UUI
     }
   }
 
+  "Update application route" should {
+    "return not found if the application does not match required statuses" in {
```
mgilgar: Where is the status that we are creating this application that belongs to the list of required status?

it/repositories/application/GeneralApplicationMongoRepositorySpec.scala
```diff
@@ -469,6 +469,41 @@ class GeneralApplicationMongoRepositorySpec extends MongoRepositorySpec with UUI
...
+      testDataRepo.createApplicationWithAllFields(UserId, AppId, "FastStream-2016",
+        additionalDoc = BSONDocument("applicationRoute" -> ApplicationRoute.Edip.toString)
+      ).futureValue
+
+      val result = repository.updateApplicationRoute(AppId, ApplicationRoute.SdipFaststream).failed.futureValue
+
+      result mustBe a[Exceptions.NotFoundException]
+    }
+
+    "update the Faststream application" in {
```
mgilgar: When the "update the Faststream application" happens? Maybe it will be clearer if you say:

update the Faststream application when condition is true

app/model/events/EmailEvents.scala
lukastymo: done

it/repositories/application/GeneralApplicationMongoRepositorySpec.scala
lukastymo: fixed

it/repositories/application/GeneralApplicationMongoRepositorySpec.scala
lukastymo: done

PR #250, opened 2016-12-02
FSET 909/1010/1011: Export a user to parity

henricook:


app/services/parity/ParityExportService.scala
```diff
@@ -0,0 +1,127 @@
...
+
+      val appDoc = applicationDoc.transform(applicationTransformer).get
+
+      val rootTransformer = (__ \ "application").json.put(appDoc) andThen
+        __.json.update(__.read[JsObject].map { o => o ++ Json.obj("token" -> parityGatewayConfig.upstreamAuthToken) })
+
+      val finalDoc = Json.toJson("{}").transform(rootTransformer)
+
+      // TODO: Validate against json schema
+      // finalDoc.get.validate()
```
kjharkin: Is the intention to validate against the schema?

app/services/parity/ParityExportService.scala
```diff
@@ -0,0 +1,127 @@
...
+  def exportApplication(applicationId: String): Future[Unit] = {
+    for {
+      exportJson <- generateExportJson(applicationId)
+      _ <- parityGatewayClient.createExport(exportJson)
+      _ <- appRepository.addProgressStatusAndUpdateAppStatus(applicationId, EXPORTED)
+    } yield ()
+  }
+
+  // scalastyle:off method.length
+  private def generateExportJson(applicationId: String): Future[JsObject] = {
```
kjharkin: (Suggestion only) Could you extract parts of this into methods/functions to make it a bit more intention revealing?

app/connectors/paritygateway/ParityGatewayClient.scala
```diff
@@ -0,0 +1,54 @@
...
+  val url = parityGatewayConfig.url
+}
+
+trait ParityGatewayClient {
+  val http: WSHttp
+  val url: String
+
+  implicit def blankedHeaderCarrier = new HeaderCarrier()
+
+  lazy val urlWithPathPrefix = s"$url/fset-parity-gateway/faststream"
```
henricook: I think it does actually, i think I tested alternatives when working on the launchpad gateway. I think parityGatewayConfig is a lazy val, so even though the object is constructed at build time it doesn't actually resolve parityGatewayConfig until this call to url. If you try to load the config object too soon i don't think Play has properly initialised and it doesn't read correctly.

PR #254, opened 2016-12-05
FSET-1020: Mark the archived applications as ARCHIVED

satishkamatam:


henricook:
Got some build fails here @satishkamatam

PR #264, opened 2016-12-06
FSET-1036: If a user completes phase3 tests, remove an expiry progress status if one was set

henricook:


henricook:
Manually tested by:

Generating a PHASE3_TESTS_EXPIRED user
Calling markAsComplete() directly (the same method used by launchpad callbacks, and the frontend redirect)

Result: User had expired removed and was in a complete status in the dashboard

app/services/onlinetesting/Phase3TestService.scala
```diff
@@ -217,20 +217,31 @@ trait Phase3TestService extends OnlineTestService with Phase3TestConcern {
...
-          updated <- phase3TestRepo.getTestGroupByToken(launchpadInviteId)
-          _ <- phase3TestRepo.updateProgressStatus(updated.applicationId, ProgressStatuses.PHASE3_TESTS_COMPLETED)
-        } yield {
-          AuditEvents.VideoInterviewCompleted(updated.applicationId) ::
-            DataStoreEvents.VideoInterviewCompleted(updated.applicationId) ::
-            Nil
+  def markAsCompleted(launchpadInviteId: String)(implicit hc: HeaderCarrier, rh: RequestHeader): Future[Unit] = {
+
+    // Launchpad only: If a user has completed unexpire them
+    def removeExpiryIfSet(applicationId: String, expired: Boolean) = expired match {
```
kjharkin: I'm not so keen on this removeExpiryIfSet method if I'm being picky. This is a form of control coupling (passing the boolean flag) that isn't necessary.  If you rename the method to removeExpiry and make the decision in the parent code whether to call it or not you will lessen the coupling.

henricook: I'd say the control coupling was a bad thing if the method wasn't an inner method of this class, it really only exists to keep the for comprehension tidy - so being control coupled was almost exactly it's purpose - i'm not sure i can see a way to rewrite this without putting an if() inside the for comp, or switching out to a big nested flatmap - what would you suggest?

kjharkin: Even in general control coupling is by no means evil - it's moderately bad at worst. I take your point about the method being a local method and I think this justifies the use of it here. 

For the record, I *would* have the 'if' condition on the method call inside the for-comprehension.

PR #266, opened 2016-12-07
FSET-1036: If a user has results sent, make sure they're unexpired

henricook:


app/services/onlinetesting/Phase3TestService.scala
```diff
@@ -267,6 +270,8 @@ trait Phase3TestService extends OnlineTestService with Phase3TestConcern {
       for {
         testGroup <- phase3TestRepo.getTestGroupByToken(launchpadInviteId)
         _ <- phase3TestRepo.updateProgressStatus(testGroup.applicationId, ProgressStatuses.PHASE3_TESTS_RESULTS_RECEIVED)
+        // Launchpad only: If results have been sent for a user, unexpire them
+        _ <- removeExpiryStatus(testGroup.applicationId)
```
ollyw: Should removeExpiryStatus be done before the progress status change, or we may end up with PHASE3_TESTS_RESULTS_RECEIVED but expired candidates if Mongo fails?

henricook: sure i can do that, i don't think i have strong feelings either way

ollyw: You may get a stronger opinion after having to investigate weird data inconsistencies like these, that can be avoided by updating in an idempotent way 😄

henricook: I don't know whether i'd rather the user ended up in results_received but expired, or got rolled back to started without any evidence that they were in expired to begin with, but with an expiry date in the past - seemed like a toss up

PR #273, opened 2016-12-12
FSET-1044 edip evaluation

benjstephenson:
Evaluate test results for EDIP phase 1.
Filter out EDIP and SDIP applications from the phase 2 and 3 test invitation schedules.


app/services/onlinetesting/ApplicationStatusCalculator.scala
```diff
@@ -27,20 +28,43 @@ import model.persisted.SchemeEvaluationResult
...
+        processResults(results, PHASE1_TESTS_PASSED, PHASE1_TESTS_FAILED)
+
       case _ => None
     }
+
+
+    applicationRoute match {
+      case ApplicationRoute.Faststream => faststreamStatusCalculator
+
+      case ApplicationRoute.Edip => edipStatusCalculator
```
satishkamatam: Might throw no match execption for Sdip, SdipFaststream

benjstephenson: Processing for those applications isn't implemented yet so I thought it would be enough to just propagate the exception up to the scheduler.  This scheduler does batch evaluation and catches the failed futures to a Failure and carries on processing the remaining applications in the batch.

PR #276, opened 2016-12-13
FSET-1041: Notify EDIP candidates who pass the online exercises

satishkamatam:


app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -372,12 +372,19 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService,
 
   override def findTestForNotification(notificationType: NotificationTestType): Future[Option[TestResultNotification]] = {
     val query = Try{ notificationType match {
-      case s: SuccessTestType => {
+      case s: SuccessTestType if s.applicationRoutes.isEmpty => {
```
benjstephenson: Do we need to add a check here for applicationRoute -> Fastastream for candidates who applied after we implemented the application route?

satishkamatam: Not required. This application route filter will just make sure the notification is sent only for specific application types.

app/model/NotificationTestType.scala
```diff
@@ -44,6 +46,12 @@ object Phase3FailedTestType
   extends FailedTestType(PHASE3_TESTS_FAILED, PHASE3_TESTS_FAILED_NOTIFIED, PHASE3_TESTS_RESULTS_RECEIVED,
       "fset_faststream_app_online_phase3_test_failed")
 
+object Phase1SuccessTestType
+  extends SuccessTestType(ApplicationStatus.PHASE1_TESTS_PASSED, PHASE1_TESTS_SUCCESS_NOTIFIED,
+    READY_FOR_EXPORT, "fset_faststream_app_online_phase1_test_success") {
+  override val applicationRoutes = List(ApplicationRoute.Edip)
+}
```
satishkamatam: For other notification types we are setting the application routes to empty means the application route filter is not considered.

PR #278, opened 2016-12-13
FSET-1051: Add missing 'result received' status when inconsistency detected

lukastymo:


app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -539,6 +552,21 @@ class GeneralApplicationMongoRepository(timeZoneService: TimeZoneService,
           "testGroups" -> "")))
         bsonCollection.findAndModify(query, updateOp).map(_.result[Candidate])
       }
+      case AddMissingPhase2ResultReceived =>
+        val query = BSONDocument("$and" -> BSONArray(
+          BSONDocument("applicationId" -> application.applicationId),
+          BSONDocument("applicationStatus" -> ApplicationStatus.PHASE2_TESTS),
+          BSONDocument(s"progress-status.${ProgressStatuses.PHASE2_TESTS_RESULTS_READY}" -> true),
+          BSONDocument(s"progress-status.${ProgressStatuses.PHASE2_TESTS_RESULTS_RECEIVED}" -> BSONDocument("$ne" -> true))
```
benjstephenson: We're re-executing part of the select query we used above to get the applications to modify.  If we're just doing this to make sure the application document hasn't changed in the meantime, should we re-run the whole query?

lukastymo: we can, but
appStatuses + progressStatuses are bare minimum to use them as guards in query. I cannot find any real situation where this should be not enough

PR #281, opened 2016-12-13
UpdateProgressStatus now ignores duplication 

ollyw:
Ignore updates which do not match the application status predicate


benjstephenson:
👍  LGTM - I'll wait for Travis and merge it down.

PR #303, opened 2017-01-05
FSET-1056: Notify SDIP-Fast Stream candidates for failed test

Lansalo:


test/services/onlinetesting/OnlineTestServiceSpec.scala
```diff
@@ -82,6 +83,42 @@ class OnlineTestServiceSpec extends UnitSpec {
...
+
+      val result = underTest.processNextTestForSdipFsNotification(FailedSdipFsTestType).futureValue
+      result mustBe unit
+
+      verify(appRepositoryMock).findTestForSdipFsNotification(FailedSdipFsTestType)
+      verifyNoMoreInteractions(appRepositoryMock)
+      verifyZeroInteractions(cdRepositoryMock, emailClientMock)
+    }
+
+    "return a successful unit future if a test to notify is found" in new OnlineTest {
```
benjstephenson: this test name doesn't reflect everything the test is doing - we're finding the application to notify and then processing the notification, not just finding the test.

PR #304, opened 2017-01-05
Sdip progression

benjstephenson:
Set passed or failed progress status for SDIP schemes


app/repositories/onlinetesting/OnlineTestEvaluationRepository.scala
```diff
@@ -155,6 +155,7 @@ class Phase1EvaluationMongoRepository()(implicit mongo: () => DB)
   }
 }
 
+
```
Lansalo: Think we can remove this line

app/repositories/onlinetesting/OnlineTestRepository.scala
```diff
@@ -171,15 +171,24 @@ trait OnlineTestRepository extends RandomSelection with ReactiveRepositoryHelper
     selectOneRandom[NotificationExpiringOnlineTest](query)
   }
 
-  def updateProgressStatus(appId: String, progressStatus: ProgressStatus): Future[Unit] = {
+  def updateProgressStatus(appId: String, progressStatus: ProgressStatus, forceApplicationStatusUpdate: Boolean = true): Future[Unit] = {
```
Lansalo: force would suggest that even when false the update is done whereas that is not the case. What do you reckon?

app/repositories/onlinetesting/Phase1TestRepository.scala
```diff
@@ -92,6 +92,22 @@ class Phase1TestMongoRepository(dateTime: DateTimeFactory)(implicit mongo: () =>
...
+    val query = BSONDocument("$and" -> BSONArray(
+      BSONDocument("applicationRoute" -> ApplicationRoute.SdipFaststream),
+      BSONDocument("testGroups.PHASE1.evaluation.result" -> BSONDocument("$elemMatch" -> BSONDocument("scheme" -> SchemeType.Sdip)))
+    ))
+    collection.find(query).one[BSONDocument] map {
+      case Some(doc) =>
+        val applicationId = doc.getAs[String]("applicationId").get
+        val userId = doc.getAs[String]("userId").get
+        val bsonPhase1 = doc.getAs[BSONDocument]("testGroups").map(_.getAs[BSONDocument](phaseName).get)
+        val phase1TestGroup = bsonPhase1.map(Phase1TestProfile.bsonHandler.read).get
```
Lansalo: Instead of a .get could we use a flatMap? Was thinking something like .flatMap(x =>  Some(Phase1TestGroupWithUserIds(applicationId, userId, x))

benjstephenson: the mongo query used means that the test profile sub-document is always present, so I thought it would be fine to use the get rather mapping an option and throwing an exception if it was not defined.

Lansalo: Makes sense. Thanks

PR #306, opened 2017-01-06
FSET-1049: Notify SDIP-Fast Stream candidates successful test

Lansalo:


app/repositories/CommonBSONDocuments.scala
```diff
@@ -106,6 +106,7 @@ trait CommonBSONDocuments extends BaseBSONReader {
           fastPassAccepted = getProgress(ProgressStatuses.FAST_PASS_ACCEPTED.key),
           withdrawn = getProgress(ProgressStatuses.WITHDRAWN.key),
           sdipFSFailed = getProgress(FailedSdipFsTestType.progressStatus),
+          sdipFSSuccessful = getProgress(SuccessfulSdipFsTestType.progressStatus),
```
satishkamatam: should we not call .key like SuccessfulSdipFsTestType.progressStatus.key

app/scheduler/onlinetesting/NotificationSdipFsTestJob.scala
```diff
@@ -32,6 +33,13 @@ object FailedSdipFsTestJob extends NotificationSdipFsTestJob {
   val config = FailedPhase1TestJobConfig
 }
 
+object SuccessfulSdipFsTestJob extends NotificationSdipFsTestJob {
+  override val service = Phase1TestService
+  override val notificationType: SuccessfulTestTypeSdipFs = SuccessfulSdipFsTestType
```
satishkamatam: Do we need this type annotation SuccessfulTestTypeSdipFs ??

app/repositories/CommonBSONDocuments.scala
benjstephenson: ProgressStatus has an implicit toString method defined so the .key calls are redundant.  I'll update the SDIP calls to make them the same though

PR #331, opened 2017-04-21
FSET-1439: Rename collections for fs2018

henricook:


henricook:
Do not merge until FSET-1424 is merged

PR #337, opened 2017-05-09
Low-risk refactors to cut down on warnings

henricook:
- Added 'silencer' (@silent) to suppress warnings we can't do anything about (because of HMRC libraries or structural changes that we're not going to make just yet, mainly use of Play.application)
- Removed unused vars, default arguments and methods, added flags to the compiler to tell us about them as warnings in future
- Moved Question/Answer out of the PersistedObjects mega file (as it's deprecated and triggers a warning) to become QuestionnaireQuestion and QuestionnaireAnswer

I've run acceptance but of course encourage you to do the same kind reviewer :-)

### DIFF
--- a/app/repositories/ReactiveRepositoryHelpers.scala
+++ b/app/repositories/ReactiveRepositoryHelpers.scala
@@ -62,7 +62,7 @@ trait ReactiveRepositoryHelpers {
     }
   }
 
-  private[this] def singleUpdateValidatorImpl(id: String, actionDesc: String, ignoreNotFound: Boolean = false,
+  private[this] def singleUpdateValidatorImpl(id: String, actionDesc: String, ignoreNotFound: Boolean,
                                               notFound: => Exception, upsert: Boolean)(result: UpdateWriteResult) = {
     if (result.ok) {
       if (result.n == 1) {
--- a/app/services/testdata/InProgressQuestionnaireStatusGenerator.scala
+++ b/app/services/testdata/InProgressQuestionnaireStatusGenerator.scala
@@ -16,13 +16,13 @@
 
 package services.testdata
 
-import model.PersistedObjects.{ PersistedAnswer, PersistedQuestion }
 import play.api.mvc.RequestHeader
 import repositories._
 import repositories.application.GeneralApplicationRepository
 import services.testdata.faker.DataFaker._
 import uk.gov.hmrc.play.http.HeaderCarrier
 import model.command.testdata.GeneratorConfig
+import model.persisted.{ QuestionnaireAnswer, QuestionnaireQuestion }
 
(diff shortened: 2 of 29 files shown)


app/repositories/ReactiveRepositoryHelpers.scala
```diff
@@ -62,7 +62,7 @@ trait ReactiveRepositoryHelpers {
     }
   }
 
-  private[this] def singleUpdateValidatorImpl(id: String, actionDesc: String, ignoreNotFound: Boolean = false,
+  private[this] def singleUpdateValidatorImpl(id: String, actionDesc: String, ignoreNotFound: Boolean,
```
benjstephenson: Why drop the default value?

app/services/testdata/InProgressQuestionnaireStatusGenerator.scala
```diff
@@ -42,61 +42,61 @@ trait InProgressQuestionnaireStatusGenerator extends ConstructiveGenerator {
...
         None
       }
     }
 
     def getHaveDegreeAnswer = {
       if (generatorConfig.isCivilServant) {
-        Some(PersistedQuestion("Do you have a degree?",
-          PersistedAnswer(Some(if (generatorConfig.hasDegree) { "Yes" } else { "No" }), None, None))
+        Some(QuestionnaireQuestion("Do you have a degree?",
+          QuestionnaireAnswer(Some(if (generatorConfig.hasDegree) { "Yes" } else { "No" }), None, None))
```
benjstephenson: Definitely a question for the wider team, but - why are we not using booleans here?

app/repositories/ReactiveRepositoryHelpers.scala
henricook: The new compiler flags said it was unused (the public methods that call this private method all have their own default  values that hide this one)

app/services/testdata/InProgressQuestionnaireStatusGenerator.scala
henricook: OH TELL ME ABOUT IT - the yes'es and no's get persisted to the database as well, legacy issues. It's possible we can change it for this year, if we talk to the boss and he approves of the work. Of course, change is risky and what's currently there works

app/repositories/ReactiveRepositoryHelpers.scala
benjstephenson: :+1:

app/services/testdata/InProgressQuestionnaireStatusGenerator.scala
benjstephenson: I remember having similar aberrations around they way the diversity questions are stored.   Let's chat about it with Fearless Leader.

PR #345, opened 2017-05-23
FSET 1354: Display the availability schedule for assessors/qacs

henricook:
- Resurrect the assessmentschedulecontroller but comment out lots of old methods. They will likely all be needed again soon, which is the reason i've not gone for a delete.
- Create a new endpoint to 'getAssessmentScheduleDatesByRegion'

### DIFF
--- a/app/controllers/AssessmentScheduleController.scala
+++ b/app/controllers/AssessmentScheduleController.scala
@@ -69,16 +69,16 @@ trait AssessmentScheduleController extends BaseController {
 
   def getAssessmentScheduleDatesByRegion(region: String): Action[AnyContent] = Action.async { implicit request =>
      acRepository.assessmentCentreCapacities.map { schedule =>
-       val toRet = schedule
-         .filter(location => location.locationName == region)
-         .flatMap(location =>
-           location.venues.flatMap(venue =>
+       val dates = schedule
+         .filter(scheduleRegion => scheduleRegion.locationName == region)
+         .flatMap(scheduleRegion =>
+           scheduleRegion.venues.flatMap(venue =>
              venue.capacityDates.map(capacityDates =>
                capacityDates.date
              )
            )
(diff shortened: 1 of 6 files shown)


app/controllers/AssessmentScheduleController.scala
```diff
@@ -0,0 +1,427 @@
...
+  // val pdRepository: PersonalDetailsRepository
+  // val cdRepository: ContactDetailsRepository
+  val emailClient: EmailClient
+  // val aaService: AssessmentCentreService
+  // val assessmentScoresService: AssessmentCentreScoresService
+
+  def getAssessmentScheduleDatesByRegion(region: String): Action[AnyContent] = Action.async { implicit request =>
+     acRepository.assessmentCentreCapacities.map { schedule =>
+       val toRet = schedule
+         .filter(location => location.locationName == region)
```
mgilgar: This is what I say in the previous PR, we use "location" although "region" may be better. We should be consistent. I am fine if we want to use region.

app/controllers/AssessmentScheduleController.scala
```diff
@@ -0,0 +1,427 @@
...
+  val auditService: AuditService
+  // val pdRepository: PersonalDetailsRepository
+  // val cdRepository: ContactDetailsRepository
+  val emailClient: EmailClient
+  // val aaService: AssessmentCentreService
+  // val assessmentScoresService: AssessmentCentreScoresService
+
+  def getAssessmentScheduleDatesByRegion(region: String): Action[AnyContent] = Action.async { implicit request =>
+     acRepository.assessmentCentreCapacities.map { schedule =>
+       val toRet = schedule
```
mgilgar: What is toRet?

henricook: A typo!

PR #353, opened 2017-06-01
FSET-1355: Logic change - Submitted is when *both* regions have been submitted, not just one or…

henricook:
… the other

### DIFF
--- a/app/repositories/AssessorAvailabilityRepository.scala
+++ b/app/repositories/AssessorAvailabilityRepository.scala
@@ -31,6 +31,7 @@ import scala.concurrent.Future
 trait AssessorAvailabilityRepository {
 
   def find(userId: String): Future[Option[AssessorAvailability]]
+
   def save(settings: AssessorAvailability): Future[Unit]
 
   def countSubmitted: Future[Int]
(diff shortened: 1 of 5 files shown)


app/services/assessoravailability/AssessorAvailabilityService.scala
```diff
@@ -29,6 +29,8 @@ object AssessorAvailabilityService extends AssessorAvailabilityService {
 trait AssessorAvailabilityService {
   val aaRepository: AssessorAvailabilityRepository
 
+  val regions = List("london", "newcastle")
```
benjstephenson: Should these come from the locations yaml datastore?

app/repositories/AssessorAvailabilityRepository.scala
```diff
@@ -55,10 +59,19 @@ class AssessorAvailabilityMongoRepository(implicit mongo: () => DB)
     val saveBson: BSONDocument = BSONDocument("$set" -> assessorAvailability)
     val insertIfNoRecordFound = true
 
-    collection.update(query, saveBson, upsert = insertIfNoRecordFound).map( _ => () )
+    collection.update(query, saveBson, upsert = insertIfNoRecordFound).map(_ => ())
   }
 
   override def countSubmitted: Future[Int] = {
-    collection.count()
+    (for {
```
mgilgar: Do we need a for comprehension if we have just one step?

PR #358, opened 2017-06-06
FSET-1542: Provide flag for turning off test assessor generation

fayimora:
This is an initial stab at making the user generation dynamic. At the moment, I've used a flag for this `?generateDefaultAssessor=false`. I won't be surprised if we will want to add the same flag for the other types of roles. 

I can either make the others dynamic too via query params OR we provide a flag to turn them all on/off OR we just do for the one flag(assessors) and add for others as the need arises.

Thoughts?

_p.s I'm making this change because it is required for a test in FSET-1542_

### DIFF
--- a/app/controllers/testdata/TestDataGeneratorController.scala
+++ b/app/controllers/testdata/TestDataGeneratorController.scala
@@ -46,8 +46,8 @@ trait TestDataGeneratorController extends BaseController {
     Ok("OK")
   }
 
-  def clearDatabase(generateDefaultRoles: Boolean) = Action.async { implicit request =>
-    TestDataGeneratorService.clearDatabase(generateDefaultRoles).map { _ =>
+  def clearDatabase(generateDefaultUsers: Boolean) = Action.async { implicit request =>
+    TestDataGeneratorService.clearDatabase(generateDefaultUsers).map { _ =>
       Ok(Json.parse("""{"message": "success"}"""))
     }
   }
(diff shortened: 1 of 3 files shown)


fayimora:
Decided to go with turning them all on/off via one flag. Commit enroute

app/controllers/testdata/TestDataGeneratorController.scala
```diff
@@ -46,8 +46,8 @@ trait TestDataGeneratorController extends BaseController {
     Ok("OK")
   }
 
-  def clearDatabase() = Action.async { implicit request =>
-    TestDataGeneratorService.clearDatabase().map { _ =>
+  def clearDatabase(generateDefaultRoles: Boolean) = Action.async { implicit request =>
```
mgilgar: I think instead of generateDefaultRoles is more like generateDefaultUsers.

fayimora: 👍

app/services/testdata/TestDataGeneratorService.scala
```diff
@@ -37,10 +37,18 @@ object TestDataGeneratorService extends TestDataGeneratorService {
 
 trait TestDataGeneratorService extends MongoDbConnection {
 
-  def clearDatabase()(implicit hc: HeaderCarrier): Future[Unit] = {
+  def clearDatabase(generateDefaultUsers: Boolean)(implicit hc: HeaderCarrier): Future[Unit] = {
     for {
       _ <- db().drop()
       _ <- AuthProviderClient.removeAllUsers()
+      _ <- generateUsers() if generateDefaultUsers
+    } yield {
```
benjstephenson: You don't need these braces here

fayimora:
@henricook I think this is ready 😉

henricook:
Love it.

PR #369, opened 2017-06-19
FSET-1618: Add  more events types

fayimora:
This PR adds more event types to the system and ensures events can be filtered properly.


it/repositories/EventsRepositorySpec.scala
```diff
@@ -53,6 +63,18 @@ class EventsRepositorySpec extends MongoRepositorySpec {
       result.head.venue mustBe VenueType.NEWCASTLE_LONGBENTON.toString
     }
 
+    "filter ALL_EVENTS in LONDON_FSAC" in {
+      repository.save(events).futureValue
+      val result = repository.fetchEvents(EventType.ALL_EVENTS, VenueType.LONDON_FSAC).futureValue
+      result.size mustBe 3
```
benjstephenson: For these kind of tests can you validate that the data retrieved contains the expected attributes?

it/repositories/EventsRepositorySpec.scala
```diff
@@ -53,6 +63,18 @@ class EventsRepositorySpec extends MongoRepositorySpec {
...
+    }
+
+    "filter FSAC in NEWCASTLE_LONGBENTON"  in {
+      repository.save(events).futureValue
+      val result = repository.fetchEvents(EventType.FSAC, VenueType.ALL_VENUES).futureValue
```
benjstephenson: This doesn't match the test name

app/controllers/EventsController.scala
```diff
@@ -19,30 +19,30 @@ package controllers
...
-import services.assessmentcentre._
+import services.events.EventsParsingService
 import uk.gov.hmrc.play.microservice.controller.BaseController
 
 import scala.concurrent.ExecutionContext.Implicits.global
 
 object EventsController extends EventsController {
   override val assessmentEventsRepository: EventsRepository = repositories.eventsRepository
-  override val assessmentCenterParsingService: AssessmentCentreParsingService = AssessmentCentreParsingService
+  override val assessmentCenterParsingService: EventsParsingService = EventsParsingService
```
kjharkin: Minor point/suggestion (i.e. no need to change it you don't want to): "override" keyword is not required for an abstract field (in parent).

app/repositories/events/EventsRepository.scala
```diff
@@ -41,8 +43,11 @@ class EventsMongoRepository(implicit mongo: () => DB)
       .map(_ => ())
   }
 
-  override def fetchEvents(eventType: EventType.Value, venue: VenueType.Value): Future[List[Event]] = {
-    val query = BSONDocument("eventType" -> eventType.toString, "venue" -> venue.toString)
+  override def fetchEvents(eventType: EventType, venue: VenueType): Future[List[Event]] = {
```
kjharkin: previous comment about "override" applies to methods too.

test/controllers/EventsControllerSpec.scala
```diff
@@ -55,19 +55,21 @@ class EventsControllerSpec extends UnitWithAppSpec {
     }
 
     "return OK with all events" in new TestFixture {
```
kjharkin: This test should be renamed perhaps as "return OK and empty event list when there are no events" or something similar?

fayimora: @kjharkin Does it need to say "and an empty List" since the test is only checking the status and not the json body?

PR #372, opened 2017-06-21
Add more stub events to the sample csv for QA

fayimora:
This adds more events to the CSV for events so QA can have something to play with.


fayimora:
@benjstephenson :shipit: 🚢

PR #373, opened 2017-06-21
FSET-1356 Rename stop the clock event to StcEvent to avoid namespace conflicts

benjstephenson:
with events schedule classes.


henricook:
Got some failing tests here @benjstephenson

app/controllers/EventsController.scala
```diff
@@ -42,6 +43,10 @@ trait EventsController extends BaseController {
     }.map(_ => Created).recover { case _ => UnprocessableEntity }
   }
 
+  def getEvent(eventId: String): Action[AnyContent] = Action.async { implicit request =>
```
henricook: What's this?

henricook: I'll accept it as it's WIP and i know you're moving on to it immediately

PR #376, opened 2017-06-23
FSET 1624 Remove parity export code and make PHASE3_TESTS_PASSED_NOTIFIED new interim terminating state. Kill a lot of old rubbish, assessment centre code we'll likely not reuse without significant changes, old TDG keys etc.

henricook:
Merge with: fset-faststream-admin-acceptance-tests/pull/74
and: fset-faststream-candidate-acceptance-tests/pull/75


conf/application.conf
```diff
@@ -186,10 +186,10 @@ microservice {
 scheduling {
   online-testing {
     send-phase1-invitation-job {
-      enabled = false
+      enabled = true
```
benjstephenson: Should these two changes be in here?

henricook: Eagle eyed

PR #383, opened 2017-07-03
FSET-1356: Add assessors to events

fayimora:


app/repositories/AllocationRepository.scala
```diff
@@ -0,0 +1,87 @@
...
+  def save(allocations: Seq[T]): Future[Unit] = {
+    val jsObjects = allocations.map(format.writes)
+    collection.bulkInsert(jsObjects.toStream, ordered = false) map (_ => ())
+  }
+
+  def delete(allocations: Seq[T]): Future[Unit] = {
+    val eventIds = allocations.map(_.eventId).distinct
+
+    val eventId = if (eventIds.size > 1) {
+      throw TooManyEventIdsException(s"The delete request contained too many event Ids [$eventIds]")
```
vglushak: What the point of having customised type exception and pass message that repeats the name of exception.
You can pass  ids as parameter for such exception, and on definition level build this message out of provided ids.

app/model/command/Allocation.scala
```diff
@@ -0,0 +1,64 @@
...
+  allocations: Seq[AssessorAllocation]
+)
+
+object AssessorAllocations {
+  implicit val assessorAllocationsFormat = Json.format[AssessorAllocations]
+
+  def apply(o: Seq[model.persisted.AssessorAllocation]): AssessorAllocations = {
+    val (opLock, eventId) = o.map( a => a.version -> a.eventId).distinct match {
+      case head :: Nil => head
+      case head :: tail => throw new Exception(s"Allocations to this event have mismatching op lock versions or event Ids ${head +: tail}")
```
vglushak: Is that the same as optimistic lock exception introduced above?

app/services/allocation/AssessorAllocationService.scala
```diff
@@ -0,0 +1,64 @@
...
+        case _ => updateExistingAllocations(existingAllocation, newAllocations).map(_ => ())
+
+      }
+    }
+  }
+
+  private def updateExistingAllocations(existingAllocation: command.AssessorAllocations,
+    newAllocations: command.AssessorAllocations): Future[Unit] = {
+
+    if (newAllocations.version == existingAllocation.version) {
```
vglushak: So, when we save allocation we need update version (so optimistic lock functionality works as expected).
I assume thats done, I just can't understand/find where it is...
Also, as for me it would be easier to have `version: Int`, and when you need to update it you just increment this number (probably there are valid reason of using String UUID).

app/repositories/AllocationRepository.scala
benjstephenson: This exception isn't specific to deleting things so I think it's still useful to be able to put information in the message string.

app/controllers/EventsController.scala
```diff
@@ -75,4 +84,22 @@ trait EventsController extends BaseController {
       case _: UnknownVenueException => BadRequest(s"$venueParam is not a valid venue")
     }
   }
+
+  def allocations(eventId: String): Action[AnyContent] = Action.async { implicit request =>
+    assessorAllocationService.getAllocations(eventId).map { allocations =>
+      if (allocations.allocations.isEmpty) {
+        NotFound
```
mgilgar: I think if there are no allocations for a particular event, it is not really an exceptional situation, it is perfectly possible. In that sense, I think returning empty list is normal. What do the others think?

app/controllers/EventsController.scala
```diff
@@ -75,4 +84,22 @@ trait EventsController extends BaseController {
...
+      } else {
+        Ok(Json.toJson(allocations))
+      }
+    }
+  }
+
+  def allocate(eventId: String): Action[JsValue] = Action.async(parse.json) { implicit request =>
```
mgilgar: I would be more explicit. I guess this is to allocate assessors to events, so given that we are in EventsController, and all methods would be event-related, the method name can be allocateAssessor

app/services/allocation/AssessorAllocationService.scala
```diff
@@ -0,0 +1,64 @@
...
+
+trait AssessorAllocationService {
+
+  def allocationRepo: AssessorAllocationMongoRepository
+
+  def getAllocations(eventId: String): Future[exchange.AssessorAllocations] = {
+    allocationRepo.allocationsForEvent(eventId).map { a => exchange.AssessorAllocations.apply(a) }
+  }
+
+  def allocate(newAllocations: command.AssessorAllocations): Future[Unit] = {
```
mgilgar: I would call it allocateAssessor to distinguish it from candidate allocation.

benjstephenson: I think AssessorAllocationService.allocate is clear enough

app/controllers/EventsController.scala
benjstephenson: agreed

app/controllers/EventsController.scala
benjstephenson: I think a 404 is synonymous with an empty list in this case.

app/controllers/EventsController.scala
```diff
@@ -31,19 +35,21 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
+  val assessorAllocationService: AssessorAllocationService = AssessorAllocationService
 }
 
 trait EventsController extends BaseController {
   def eventsService: EventsService
-  def locationsAndVenues: LocationsWithVenuesRepository
+  def locationsAndVenuesRepository: LocationsWithVenuesRepository
+  def assessorAllocationService: AssessorAllocationService
 
   def venuesForEvents: Action[AnyContent] = Action.async { implicit request =>
```
kjharkin: I can see locations and venues being used by other things and hence should probably be exposed through a custom endpoint/controller. I think it is fine to leave it here for now until we know for sure.

conf/locations-and-venues.yaml
```diff
@@ -1,10 +1,13 @@
 London:
-  - London FSAC:
+  - LONDON_FSAC:
       description: Bush House
```
vglushak: [note]
It's not really description.. it's display value (or label) that we will show on UI. We can't show LONDON_FSAC.
I suggest to change it in scope of my story FSET-1553 if you're ok with it.

kjharkin:
Looks good - will merge once the other related PRs are reviewed (and the tests have run)

PR #385, opened 2017-07-04
Fset-1630 updated schemes.yaml and added few additional events schedu…

emailvacharya:
…les to test in QA


andrewrembrandt:
Minor feedback via chat: 
The virtual on the last row should have an uppercase V?

vijay Acharya [11:11 AM] 
ah .. is it

[11:11] 
ok.. let me that change

Andrew Rembrandt [11:11 AM] 
And should telephone not have a capital T?

[11:11] 
(row 5)

new messages
vijay Acharya [11:12 AM] 
ok

PR #389, opened 2017-07-06
FSET 1428: Backup report for FS2017

henricook:


app/repositories/application/PreviousYearCandidatesDetailsRepository.scala
```diff
@@ -0,0 +1,498 @@
...
+class PreviousYearCandidatesDetailsMongoRepository(implicit mongo: () => DB) extends PreviousYearCandidatesDetailsRepository with CommonBSONDocuments {
+  import config.MicroserviceAppConfig._
+
+  val applicationDetailsCollection = mongo().collection[JSONCollection](CollectionNames.APPLICATION)
+
+  val contactDetailsCollection = mongo().collection[JSONCollection](CollectionNames.CONTACT_DETAILS)
+
+  val questionnaireCollection = mongo().collection[JSONCollection](CollectionNames.QUESTIONNAIRE)
+
+  private def optYes = Some("Yes")
```
benjstephenson: Would an implicit remove the need to check booleans and return these values everywhere?

henricook: I'm not a huge fan of implicits on primitives

PR #393, opened 2017-07-09
Refactor skill path parameters in ApplicationClient

benjstephenson:
Accept skill types in routes as a list from query params rather than a comma separated string path segment.

### DIFF
--- a/conf/app.routes
+++ b/conf/app.routes
@@ -127,8 +127,8 @@ GET         /events/:venue/:eventType                                    control
 PUT         /events/:eventId/allocate                                    controllers.EventsController.allocateAssessor(eventId: String)
 
 # Day aggregate events
-GET         /day-aggregate-events/skills/:skillTypes                     controllers.DayAggregateEventController.findBySkillTypes(skillTypes: String)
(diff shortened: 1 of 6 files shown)


conf/app.routes
```diff
@@ -127,8 +127,8 @@ GET         /events/:venue/:eventType                                    control
 PUT         /events/:eventId/allocate                                    controllers.EventsController.allocateAssessor(eventId: String)
 
 # Day aggregate events
-GET         /day-aggregate-events/skills/:skillTypes                     controllers.DayAggregateEventController.findBySkillTypes(skillTypes: String)
-GET         /day-aggregate-events/location/:location/skills/:skillTypes  controllers.DayAggregateEventController.findBySkillTypesAndLocation(skillTypes: String, location: String)
+GET         /day-aggregate-events                                        controllers.DayAggregateEventController.findBySkillTypes(skills: List[model.persisted.eventschedules.SkillType.SkillType])
```
kjharkin: Not sure of you've refactored this in your related story but the package naming of SkillType looks wrong.

benjstephenson: I had started a bunch of small fixes/refactors in that other PR and decided to move them out to separate ones to make it easier to review.  I'll check the package naming.

benjstephenson: This is the right package for the enum - I think we need to look at moving to AssessorSkill case class in some way as we've introduced that last week.

PR #396, opened 2017-07-10
Use the skill type enum when creating the assessor users in the TDG so

benjstephenson:
the skill values are correct


app/services/testdata/candidate/RegisteredStatusGenerator.scala
```diff
@@ -82,7 +83,8 @@ trait RegisteredStatusGenerator extends BaseGenerator {
     val assessorRoles = List(AuthProviderClient.AssessorRole, AuthProviderClient.QacRole)
     userFuture.flatMap {
       case user if assessorRoles.contains(role) =>
-        assessorGenerator.createAssessor(user.userId, AssessorData(List("assessor", "qac", "sifter"),
```
vglushak: it's already fixed in 1553

PR #401, opened 2017-07-11
Fset 1633 - backend for sifting candidates

WarFox:


app/repositories/sifting/SiftingRepository.scala
```diff
@@ -0,0 +1,95 @@
...
+class SiftingMongoRepository()(implicit mongo: () => DB)
+  extends ReactiveRepository[CreateApplicationRequest, BSONObjectID](CollectionNames.APPLICATION, mongo,
+    Commands.Implicits.createApplicationRequestFormat,
+    ReactiveMongoFormats.objectIdFormats) with SiftingRepository with RandomSelection with CommonBSONDocuments
+    with GeneralApplicationRepoBSONReader with ReactiveRepositoryHelpers {
+
+
+  /**
+    * TODO: implement all criterias
+    * Criteria:
```
fayimora: Might want to get rid of your todo

it/repositories/sifting/SiftingRepositorySpec.scala
```diff
@@ -0,0 +1,88 @@
...
+
+    "sift candidate as Passed" in {
+      createSiftEligibleCandidates(UserId, AppId)
+      repository.siftCandidate(AppId, SchemeEvaluationResult(Commercial, "Green")).futureValue
+      val candidates = repository.findApplicationsReadyForSifting(Commercial).futureValue
+      candidates.size mustBe 0
+    }
+
+    /*
+    //    TODO: implement this case
```
fayimora: This should go too

app/controllers/SiftingController.scala
```diff
@@ -0,0 +1,60 @@
...
+  val siftAppRepository = siftingRepository
+}
+
+trait SiftingController extends BaseController {
+
+  import Implicits._
+
+  val siftAppRepository: SiftingRepository
+
+  def findSiftingEligible(schemeId: String): Action[AnyContent] = Action.async { implicit request =>
```
mgilgar: What are we finding here? Applications? I would call this method exactly as the one in repo findApplicationsReadyForSifting.

app/controllers/SiftingController.scala
```diff
@@ -0,0 +1,60 @@
...
+    siftAppRepository.findApplicationsReadyForSifting(SchemeId(schemeId)).map { candidates =>
+      Ok(Json.toJson(candidates))
+    }
+  }
+
+  def fromPassMark(s: String): EvaluationResults.Result = s match {
```
mgilgar: I think this method should be private, I think the only methods that should be public in controller are the ones exposed in app.routes.

app/controllers/SiftingController.scala
```diff
@@ -0,0 +1,60 @@
...
+    case "Pass" => EvaluationResults.Green
+    case "Fail" => EvaluationResults.Red
+    case _ => sys.error(s"Unsupported evaluation result $s for sifting")
+  }
+
+
+  def submitSifting: Action[JsValue] = Action.async(parse.json) { implicit request =>
```
mgilgar: I am not sure on this suggestion, but I would just say:
def siftCandidate...

BTW do you sift a Candidate or an Application? Not sure if in FastStream a candidate can have different Applications.

it/repositories/sifting/SiftingRepositorySpec.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  def repository = new SiftingMongoRepository()
+
+  private val userId = "123"
+
+  private val Commercial: SchemeId = SchemeId("Commercial")
+  private val European: SchemeId = SchemeId("European")
+  private val Sdip: SchemeId = SchemeId("Sdip")
+
+  "Sifting repository" should {
```
mgilgar: There are different styles in the source code but I prefer the structure it in this way, forget about: "Sifting repository" should {
because we are in SiftingRepository already and use something like:
"find eligible candidates" should {
   "return candidates in a particular state when whatever condition" in { ...

test/controllers/SiftingControllerSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+class SiftingControllerSpec extends UnitWithAppSpec {
+  val mockSiftingRepo = mock[SiftingRepository]
+
+  val controller = new SiftingController {
+    override val siftAppRepository: SiftingRepository = mockSiftingRepo
+  }
+
+  private val Commercial = SchemeId("Commercial")
+
+  "find sifting eligible" should {
```
mgilgar: I like this style, I think the repo should be in the same way as I point out above.

it/repositories/sifting/SiftingRepositorySpec.scala
WarFox: yes! this is what I mentioned during standup! I noticed it only after raising pull request. Will all TODOs and clear them 👍

app/controllers/SiftingController.scala
WarFox: I thought a candidate can have multiple applications. We are sifting a candidate's application for the selected scheme. Probably we should call this method `siftCandidateApplication`

PR #402, opened 2017-07-11
Fix assessor availability tdg endpoint

mgilgar:
The create assessor user endpoint in Test Data Generator could not create the availability which has been fixed in this PR.

### DIFF
--- a/app/controllers/testdata/TestDataGeneratorController.scala
+++ b/app/controllers/testdata/TestDataGeneratorController.scala
@@ -146,7 +146,8 @@ trait TestDataGeneratorController extends BaseController {
       role = Some("assessor"),
       phone = Some("123456789"),
       assessor = Some(AssessorRequest(
-        skills = Some(List("assessor", "qac")),
+        skills = Some(List("ASSESSOR", "QUALITY_ASSURANCE_COORDINATOR")),
+        sifterSchemes = Some(List(SchemeId("GovernmentEconomicsService"), SchemeId("ProjectDelivery"), SchemeId("Sdip"))),
         civilServant = Some(true),
         availability = Some(List(
           AssessorAvailabilityRequest("London", LocalDate.now()),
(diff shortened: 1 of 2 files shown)


app/controllers/testdata/TestDataGeneratorController.scala
```diff
@@ -146,8 +146,7 @@ trait TestDataGeneratorController extends BaseController {
       role = Some("assessor"),
       phone = Some("123456789"),
       assessor = Some(AssessorRequest(
-        skills = Some(List("ASSESSOR", "QUALITY_ASSURANCE_COORDINATOR")),
-        sifterSchemes = Some(List(SchemeId("GovernmentEconomicsService"), SchemeId("ProjectDelivery"), SchemeId("Sdip"))),
```
henricook: This looks like an accidental regression, is it?

mgilgar: Yes, it is. It is weird.

henricook: Just a merge that resolved wrong? Do you want to correct?

app/controllers/testdata/TestDataGeneratorController.scala
```diff
@@ -165,7 +164,7 @@ trait TestDataGeneratorController extends BaseController {
       eventType = Some(EventType.FSAC),
       description = Some("PDFS FSB"),
       location = Some("London"),
-      venue = Some("LONDON_FSAC"),
+      venue = Some("London venue 1"),
```
henricook: This line probably also a merge issue?

mgilgar: Yes. It should be fixed now

app/controllers/testdata/TestDataGeneratorController.scala
mgilgar: Yes, it should be fixed now.

PR #403, opened 2017-07-12
Fset 1674 - Event/Sessions config spike

andrewrembrandt:
Moved the Events definition into Yaml and provided nested sessions (within the Events)

### DIFF
--- a/app/repositories/events/EventsConfigRepository.scala
+++ b/app/repositories/events/EventsConfigRepository.scala
@@ -89,12 +89,12 @@ trait EventsConfigRepository {
     val eventsConfig = yamlAst.convertTo[List[EventConfig]]
 
     // Force all 'types' to be upper case and replace hyphens with underscores
-    eventsConfig.map(configItem => configItem.copy(
+    val massagedEventsConfig = eventsConfig.map(configItem => configItem.copy(
       eventType = configItem.eventType.replaceAll("\\s|-", "_").toUpperCase,
       skillRequirements = configItem.skillRequirements.map {
         case (skillName, numStaffRequired) => (skillName.replaceAll("\\s|-", "_").toUpperCase, numStaffRequired)}))
 
-    FutureEx.traverseSerial(eventsConfig) { case configItem =>
+    FutureEx.traverseSerial(massagedEventsConfig) { case configItem =>
       val eventItemFuture = for {
         location <- locationsWithVenuesRepo.location(configItem.location)
         venue <- locationsWithVenuesRepo.venue(configItem.venue)
--- a/app/repositories/events/EventsConfigRepository.scala
+++ b/app/repositories/events/EventsConfigRepository.scala
@@ -16,6 +16,7 @@
 
 package repositories.events
 
+import common.FutureEx
 import config.MicroserviceAppConfig
 import factories.UUIDFactory
 import model.persisted.eventschedules._
@@ -45,24 +46,6 @@ case class EventConfig(
                   sessions: List[Session]
                 )
 
-object EventConfig {
-  implicit def configToEvent(c: EventConfig): Event = {
-    Event(UUIDFactory.generateUUID(),
-      EventType.withName(c.eventType.replaceAll("\\s|-", "_").toUpperCase),
-      c.description,
-      Location(c.location.toUpperCase),
-      Venue(c.venue.toUpperCase, ""),
-      c.date,
-      c.capacity,
-      c.minViableAttendees,
-      c.attendeeSafetyMargin,
-      c.startTime,
-      c.endTime,
-      c.skillRequirements,
-      c.sessions)
-  }
-}
-
 object EventConfigProtocol extends DefaultYamlProtocol {
   implicit object LocalDateYamlFormat extends YamlFormat[LocalDate] {
     def write(jodaDate: LocalDate) = YamlDate(jodaDate.toDateTimeAtStartOfDay)
@@ -90,6 +73,8 @@ object EventConfigProtocol extends DefaultYamlProtocol {
 }
 
 trait EventsConfigRepository {
+  def locationsWithVenuesRepo: LocationsWithVenuesRepository
+
   import play.api.Play.current
 
   protected def rawConfig = {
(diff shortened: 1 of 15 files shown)


app/services/events/EventsService.scala
```diff
@@ -36,13 +37,22 @@ trait EventsService {
...
+
+  def saveAssessmentEventsFromDeprecatedCsv(): Future[Unit] = {
     eventFileParsingService.processCentres().flatMap { events =>
       Logger.debug("Events have been processed!")
       eventsRepo.save(events)
     }
   }
 
+  def saveAssessmentEvents(): Future[Unit] = {
+    eventsConfigRepo.events.flatMap { events =>
```
vglushak: I know, it's not part of spike, but probably if you have time, it would be good to make more robust solution.
There is possibility to save event with invalid location/venue/event type (yes, you wrap it in object, but it never check if such value is in config). 
For example there is nice logic of creating venue from name based on config in `EventsController`: 
`locationsAndVenuesRepository.venue(venueParam).flatMap { venue =>`
The same validation can be done for most enum-like fields.

app/repositories/events/EventsConfigRepository.scala
```diff
@@ -0,0 +1,112 @@
...
+                  description: String,
+                  location: String,
+                  venue: String,
+                  date: LocalDate,
+                  capacity: Int,
+                  minViableAttendees: Int,
+                  attendeeSafetyMargin: Int,
+                  startTime: LocalTime,
+                  endTime: LocalTime,
+                  resourceRequirements: Map[String, Int],
```
vglushak: i think it's skillRequirements in Event. Is that different naming done in purpose?

app/repositories/events/EventsConfigRepository.scala
```diff
@@ -0,0 +1,112 @@
...
+      c.sessions)
+  }
+}
+
+object EventConfigProtocol extends DefaultYamlProtocol {
+  implicit object LocalDateYamlFormat extends YamlFormat[LocalDate] {
+    def write(x: LocalDate) = YamlDate(x.toDateTimeAtStartOfDay)
+    def read(value: YamlValue) = value match {
+      case YamlDate(x) => x.toLocalDate
+      case x => deserializationError("Expected Date as YamlDate, but got " + x)
```
vglushak: string interpolation ? `$x`

app/services/events/EventsService.scala
andrewrembrandt: Will chat with Wamiq, but valid feedback, will start on that now

test/services/events/EventsConfigRepositorySpec.scala
```diff
@@ -0,0 +1,124 @@
...
+import org.joda.time.{ LocalDate, LocalTime }
+import org.scalatest.Matchers
+import org.scalatest.time.{ Millis, Seconds, Span }
+import org.scalatestplus.play.PlaySpec
+import uk.gov.hmrc.play.test.UnitSpec
+
+class EventsConfigRepositorySpec extends UnitSpec with Matchers with ScalaFutures {
+  "events" must {
+    "successfully parse the config" in {
+      val input = """- eventType: FSAC
```
vglushak: I believe it would be useful to have test that load yams file (if possible) and parse it (at least to make sure that file we submitted is correct).

app/repositories/events/EventsConfigRepository.scala
```diff
@@ -0,0 +1,112 @@
...
+      c.startTime,
+      c.endTime,
+      c.resourceRequirements,
+      c.sessions)
+  }
+}
+
+object EventConfigProtocol extends DefaultYamlProtocol {
+  implicit object LocalDateYamlFormat extends YamlFormat[LocalDate] {
+    def write(x: LocalDate) = YamlDate(x.toDateTimeAtStartOfDay)
```
mgilgar: Instead of x I would write something else like localDate or date

app/repositories/events/EventsConfigRepository.scala
```diff
@@ -0,0 +1,112 @@
...
+                  sessions: List[Session]
+                )
+
+object EventConfig {
+  implicit def configToEvent(c: EventConfig): Event = {
+    Event(UUIDFactory.generateUUID(),
+      EventType.withName(c.eventType),
+      c.description,
+      Location(c.location),
+      Venue(c.venue, ""),
```
vglushak: Second parameter can be picked up from LocationAndVenue repository.

app/repositories/events/EventsConfigRepository.scala
```diff
@@ -0,0 +1,112 @@
...
+      c.resourceRequirements,
+      c.sessions)
+  }
+}
+
+object EventConfigProtocol extends DefaultYamlProtocol {
+  implicit object LocalDateYamlFormat extends YamlFormat[LocalDate] {
+    def write(x: LocalDate) = YamlDate(x.toDateTimeAtStartOfDay)
+    def read(value: YamlValue) = value match {
+      case YamlDate(x) => x.toLocalDate
```
mgilgar: Instead of x I would write something more descriptive. Actually I am not sure what is x unless I go inside YamlDate declaration.

app/repositories/events/EventsConfigRepository.scala
```diff
@@ -0,0 +1,112 @@
...
+      case x => deserializationError("Expected Date as YamlDate, but got " + x)
+    }
+  }
+
+  implicit object LocalTimeYamlFormat extends YamlFormat[LocalTime] {
+    def write(x: LocalTime) = YamlString(x.toString("HH:mm"))
```
mgilgar: x

app/repositories/events/EventsConfigRepository.scala
```diff
@@ -0,0 +1,112 @@
...
+    def read(value: YamlValue) = value match {
+      case YamlString(s) => DateTimeFormat.forPattern("HH:mm").parseLocalTime(s)
+      case YamlNumber(s) => {
+        val hour = s.toInt / 60
+        val minute = s % 60
+        DateTimeFormat.forPattern("HH:mm").parseLocalTime(s"$hour:$minute")
+      }
+      case x => deserializationError("Expected Time as YamlString/YamlNumber, but got " + x)
```
mgilgar: What is x here? Do you mean _?

app/repositories/events/EventsConfigRepository.scala
```diff
@@ -0,0 +1,112 @@
...
+    }
+  }
+
+  implicit val sessionFormat = yamlFormat3((a: String, b: LocalTime, c: LocalTime) => Session(a,b,c))
+  implicit val eventFormat = yamlFormat12((a: String, b: String, c: String, d: String,
+                                           e: LocalDate, f: Int, g: Int, h: Int,
+                                           i: LocalTime, j: LocalTime, k: Map[String, Int], l: List[Session]) =>
+    EventConfig(a,b,c,d,e,f,g,h,i,j,k,l))
```
mgilgar: What is a, b, c, d, e, f, g, h, i, j, k, l?

app/repositories/events/EventsConfigRepository.scala
```diff
@@ -0,0 +1,112 @@
...
+      case YamlNumber(s) => {
+        val hour = s.toInt / 60
+        val minute = s % 60
+        DateTimeFormat.forPattern("HH:mm").parseLocalTime(s"$hour:$minute")
+      }
+      case x => deserializationError("Expected Time as YamlString/YamlNumber, but got " + x)
+    }
+  }
+
+  implicit val sessionFormat = yamlFormat3((a: String, b: LocalTime, c: LocalTime) => Session(a,b,c))
```
mgilgar: What is a, b, c?

app/repositories/events/EventsConfigRepository.scala
```diff
@@ -0,0 +1,112 @@
...
+
+  protected def rawConfig = {
+    val input = managed(Play.application.resourceAsStream(MicroserviceAppConfig.eventsConfig.yamlFilePath).get)
+    input.acquireAndGet(stream => Source.fromInputStream(stream).mkString)
+  }
+
+  lazy val events = Future {
+    import EventConfigProtocol._
+
+    val yamlAst = rawConfig.parseYaml
```
mgilgar: What is Ast in yamlAst?

app/services/testdata/faker/DataFaker.scala
```diff
@@ -772,6 +772,8 @@ object DataFaker {
           skillSelected -> numberOfPeopleWithSkillsRequired
         }.toMap
       }
+
+      def sessions = randList(List(Session("First", startTime, startTime.plusHours(1))),1)
```
mgilgar: Not sure if it is necessary at this moment but you can combine randList with number like this:

def sessions = randList(List(Session("First", startTime, startTime.plusHours(1))),number(2))

where 2 means to number of maximum of sessions per day.

There is a method called number(Int) in DataFaker.Random object that generates a random integer.

app/repositories/events/EventsConfigRepository.scala
andrewrembrandt: Just updated back to skillRequirements

app/repositories/events/EventsConfigRepository.scala
andrewrembrandt: https://en.wikipedia.org/wiki/Abstract_syntax_tree

kjharkin:
Looks good - will merge once local tests complete

PR #404, opened 2017-07-12
FSET-1649 Notify candidate of invitation to FSAC event

vglushak:


app/connectors/EmailClient.scala
```diff
@@ -209,6 +209,18 @@ trait EmailClient extends WSHttp {
   def sendApplicationExtendedToSdip(to: String, name: String)(implicit hc: HeaderCarrier): Future[Unit] =
     sendEmail(to, "fset_faststream_app_converted_to_sdip_confirmation", Map("name" -> name))
 
+  def sendCandidateInvitationRequestToFSAC(to: String, name: String,
```
mgilgar: Shouldn't this be sendCandidateConfirmationRequestToFSAC?

app/model/stc/EmailEvents.scala
```diff
@@ -32,4 +32,9 @@ object EmailEvents {
   case class AdjustmentsConfirmed(to: String, name: String, etrayAdjustments: String, videoAdjustments: String) extends EmailEvent
   case class AdjustmentsChanged(to: String, name: String, etrayAdjustments: String, videoAdjustments: String) extends EmailEvent
   case class ApplicationConvertedToSdip(to: String, name: String) extends EmailEvent
+
+  case class CandidateAllocationConfirmed(to: String, name: String, eventDate: String, eventTime: String) extends EmailEvent
+  case class CandidateAllocationRequestConfirmation(to: String, name: String, eventDate: String, eventTime: String,
```
mgilgar: I think should be verbs in the past. Maybe CandidateAllocationConfirmationRequested.

app/services/allocation/AssessorAllocationService.scala
```diff
@@ -16,31 +16,50 @@
...
+import uk.gov.hmrc.play.http.HeaderCarrier
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
 object AssessorAllocationService extends AssessorAllocationService {
-  def allocationRepo = repositories.assessorAllocationRepository
-  override val eventsService = EventsService
-  def candidateAllocationRepo = repositories.candidateAllocationRepository
+  val allocationRepo: AssessorAllocationMongoRepository = repositories.assessorAllocationRepository
```
mgilgar: To distinguish clearly from candidateAllocationRepo, I would call it assessorAllocationRepo

app/services/allocation/AssessorAllocationService.scala
```diff
@@ -16,31 +16,50 @@
...
+  val candidateAllocationRepo: CandidateAllocationMongoRepository = repositories.candidateAllocationRepository
+  val applicationRepo: GeneralApplicationRepository = repositories.applicationRepository
+  val eventsService = EventsService
```
mgilgar: I would put all services together and all clients together.

app/services/allocation/AssessorAllocationService.scala
```diff
@@ -56,27 +75,75 @@ trait AssessorAllocationService {
...
-  def allocateCandidates(newAllocations: command.CandidateAllocations): Future[Unit] = {
-    getCandidateAllocations(newAllocations.eventId).flatMap { existingAllocation =>
-      existingAllocation.allocations match {
-        case Nil => candidateAllocationRepo.save(persisted.CandidateAllocation.fromCommand(newAllocations)).map(_ => ())
-        case _ => updateExistingAllocations(existingAllocation, newAllocations).map(_ => ())
+  private val dateFormat = "dd MMMM YYYY"
+  private val timeFormat = "HH:mma"
+
+
+  def allocateCandidates(newAllocations: command.CandidateAllocations)(implicit hc: HeaderCarrier, rh: RequestHeader): Future[Unit] = {
```
mgilgar: Why do you need the allocateCandidates method in AssessorAllocationService.scala file? I think we need a separate CandidateAllocationService in the same way we have a candidateAllocationRepo and assessorAllocationRepo.

test/services/allocation/AssessorAllocationServiceSpec.scala
```diff
@@ -80,14 +84,35 @@ class AssessorAllocationServiceSpec extends BaseServiceSpec {
     }
   }
 
+  "Allocate candidate" must {
```
mgilgar: If you decide to move allocate candidates to its own service, you will have to create a separate spec I guess for this test.

app/connectors/EmailClient.scala
vglushak: yes.

app/model/stc/EmailEvents.scala
vglushak: yes.

app/services/allocation/AssessorAllocationService.scala
vglushak: ok.

app/services/allocation/AssessorAllocationService.scala
vglushak: Well, yes, but:
- out of scope of this task.
- taking into account current test coverage I would prefer not to do unrelated refactorings.

mgilgar: It seems this is new code: in this case it is not refactoring. Or do you move it inside the file?

vglushak: I've added implicit arguments set, but method already exists.

kjharkin: This should be moved but we can do so when we revisit the allocations.

app/services/allocation/AssessorAllocationService.scala
```diff
@@ -16,67 +16,133 @@
...
+      val eventTime = event.startTime.toString(timeFormat)
+      val deadlineDateTime = event.date.minusDays(10).toString(dateFormat)
+
+
+      getCandidateAllocations(newAllocations.eventId).flatMap { existingAllocation =>
+        existingAllocation.allocations match {
+          case Nil =>
+            Future.sequence(newAllocations.allocations.map(sendCandidateEmail(_, eventDate, eventTime, deadlineDateTime)))
+              .flatMap { _ =>
+                candidateAllocationRepo.save(persisted.CandidateAllocation.fromCommand(newAllocations)).map(_ => ())
```
kjharkin: I think the order here is wrong - we should save the allocations before sending the email, otherwise if things go wrong we may invite a candidate without having an allocation for them - could be embarrassing.

app/services/allocation/AssessorAllocationService.scala
```diff
@@ -16,67 +16,133 @@
...
+              }
+          case _ =>
+            val existingIds = existingAllocation.allocations.map(_.id)
+            Future.sequence(
+              newAllocations.allocations
+                .filter(alloc => !existingIds.contains(alloc.id))
+                .map(sendCandidateEmail(_, eventDate, eventTime, deadlineDateTime))
+            ).flatMap { _ =>
+              updateExistingAllocations(existingAllocation, newAllocations).map(_ => ())
```
kjharkin: Same here

app/services/allocation/AssessorAllocationService.scala
```diff
@@ -16,67 +16,133 @@
...
+          val res = authProviderClient.findByUserIds(Seq(candidate.userId)).map { candidates =>
+            candidates.map { candidate =>
+              candidateAllocation.status match {
+                case AllocationStatuses.UNCONFIRMED =>
+                  CandidateAllocationConfirmationRequest(candidate.email, candidate.name, eventDate, eventTime, deadlineDateTime)
+                case AllocationStatuses.CONFIRMED =>
+                  CandidateAllocationConfirmed(candidate.email, candidate.name, eventDate, eventTime)
+              }
+            }
+          } recover { case _ => throw new RuntimeException("Was not able to retrieve user details.") }
```
kjharkin: whose user details?

app/services/allocation/AssessorAllocationService.scala
```diff
@@ -16,67 +16,133 @@
...
+          res.asInstanceOf[Future[StcEvents]]
+        }
+      case None => throw new RuntimeException("Can not find user application")
```
kjharkin: which user application?

PR #405, opened 2017-07-12
Do not create availability by default, only if status is relevant

vglushak:
Basically this is PR to fix acceptance tests.

@mgilgar  please review.

### DIFF
--- a/app/model/testdata/CreateAdminData.scala
+++ b/app/model/testdata/CreateAdminData.scala
@@ -19,6 +19,7 @@ package model.testdata
 import model.SchemeId
 import model.command.testdata.CreateAdminRequest.CreateAdminRequest
 import model.exchange.AssessorAvailability
+import model.persisted.assessor.AssessorStatus
 import org.joda.time.LocalDate
 import play.api.libs.json.{ Json, OFormat }
 import services.testdata.faker.DataFaker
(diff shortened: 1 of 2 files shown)


app/model/testdata/CreateAdminData.scala
```diff
@@ -44,12 +45,17 @@ object CreateAdminData {
...
-        val availability: Option[List[AssessorAvailability]] = createRequest.assessor.flatMap(assessorRequest => {
+        val availability1: Option[List[AssessorAvailability]] = createRequest.assessor.flatMap(assessorRequest => {
             assessorRequest.availability.map { assessorAvailabilityRequests => {
               assessorAvailabilityRequests.map { assessorAvailabilityRequest => {
                 AssessorAvailability.apply(assessorAvailabilityRequest)
               }
-            }}}}).orElse(DataFaker.Random.Assessor.availability)
+            }}}})
+        val availability = createRequest.assessor match {
+          case Some(assessorRequest) if assessorRequest.status == AssessorStatus.AVAILABILITIES_SUBMITTED && availability1.isEmpty =>
```
mgilgar: The way we handle the different statuses in the case of candidates is by creating another status generator class, i.e., so we have CreatedStatusGenerator, InProgressPersonalDetailsStatusGenerator, InProgressAssistanceDetailsStatusGenerator. 

The generation of admin and assessor is thought to work in the same way, although is not there yet, because we did not have statuses before.

Maybe we can have a discussion in person about this.

so in theory this logic should not be in CreateAdminData but

vglushak:
Tests fixed in different way.

PR #408, opened 2017-07-13
Add date field to exchange object

benjstephenson:


app/model/exchange/EventAllocations.scala
```diff
@@ -27,9 +28,12 @@ object EventAssessorAllocationsSummaryPerSkill {
   implicit val format = Json.format[EventAssessorAllocationsSummaryPerSkill]
 }
 
-case class EventWithAllocationsSummary(event: Event,
-                                       confirmedCandidates: Int,
-                                       assessorAllocations: List[EventAssessorAllocationsSummaryPerSkill])
+case class EventWithAllocationsSummary(
+  date: LocalDate,
```
vglushak: I do not think it's needed. You can always access date from event. (see my comment at frontend PR)

benjstephenson: It should be refactored to a flatter structure.  I need it here to generically handle it in the calendar builder.

PR #416, opened 2017-07-19
Fset 1633a PHASE3_TESTS_PASSED_NOTIFIED

WarFox:
Sift candidates from PHASE3_TESTS_PASSED_NOTIFIED

### DIFF
--- a/app/repositories/sifting/SiftingRepository.scala
+++ b/app/repositories/sifting/SiftingRepository.scala
@@ -19,8 +19,9 @@ package repositories.sifting
 import model.Commands.{ Candidate, CreateApplicationRequest }
 import model.EvaluationResults.Green
 import model.Exceptions.ApplicationNotFound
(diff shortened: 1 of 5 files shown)


app/repositories/sifting/SiftingRepository.scala
```diff
@@ -56,8 +57,8 @@ class SiftingMongoRepository()(implicit mongo: () => DB)
     )
 
     val query = BSONDocument("$and" -> BSONArray(
-      BSONDocument(s"applicationStatus" -> ApplicationStatus.PHASE3_TESTS_PASSED),
-      BSONDocument(s"progress-status.${ApplicationStatus.PHASE3_TESTS_PASSED}" -> true),
+      BSONDocument(s"applicationStatus" -> ApplicationStatus.PHASE3_TESTS_PASSED_NOTIFIED),
```
vglushak: Why not query for both statuses?

PR #419, opened 2017-07-20
FSET-1678 Update look up by postcode file and default to Newcastle

WarFox:
**FSET-1678 Update look up by postcode file and default to Newcastle**


henricook:
LGTM :+1:

PR #421, opened 2017-07-20
Change the 'id' field to 'userId' on assessor allocations

henricook:
I was working with this case class on a branch and kept getting caught out by 'id' - is it an assessor allocation id? a user id? some other kind of id? So i've changed it to make it clearer

@fayimora @vglushak - i think this is your area of responsibility now :-) do you think i've got everything and it's ok?

### DIFF
--- a/app/model/command/Allocation.scala
+++ b/app/model/command/Allocation.scala
@@ -22,7 +22,6 @@ import model.exchange.AssessorSkill
 import play.api.libs.json.{ Json, OFormat }
 
 trait Allocation {
-  def userId: String
   def status: AllocationStatus
 }
 
@@ -58,7 +57,7 @@ object AssessorAllocations {
 
     AssessorAllocations(opLock, eventId, o.map { a =>
       val allocatedSkill = AssessorSkill.SkillMap(a.allocatedAs)
-      AssessorAllocation(a.id, a.status, allocatedSkill)
+      AssessorAllocation(a.userId, a.status, allocatedSkill)
     })
   }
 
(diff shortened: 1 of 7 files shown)


vglushak:
👍 
But for CandidateAllocation it's applicationId, rather then user.

fayimora:
I'm happy with that change. You will need to propagate it to the templates though as they are used there and I dont see any templates in your diff.

henricook:
Exactly why I asked! Templates in admin frontend only I guess right? We use that model for exchange? I'll take a look

henricook:
Oh i see, to @vglushak - the id means a different thing in CandidateAllocation - yipes! Will have a think about it, maybe the base type isn't necessary

benjstephenson:
I would say the base type isn't necessary anymore as the candidate
allocation will change to include all the session related info.

app/model/command/Allocation.scala
```diff
@@ -22,14 +22,13 @@ import model.exchange.AssessorSkill
...
   def status: AllocationStatus
 }
 
 case class AssessorAllocation(
-  id: String,
-  status: AllocationStatus,
-  allocatedAs: AssessorSkill
+                               userId: String,
+                               status: AllocationStatus,
+                               allocatedAs: AssessorSkill
```
vglushak: https://stackoverflow.com/questions/26880677/intellij-scala-plugins-case-class-indentation-is-absurd

vglushak:
@benjstephenson the way it's organised it does not need to include event/session id (those two parameters passed outside json). We might want to rework it in future, but it works on session level already.
-
LGTM.

henricook:
My solution to this got too ugly, so i'm abandoning. Refactoring the very generic repository with a different ID key was a bit rough.

PR #423, opened 2017-07-23
FSET-1659: Create Event and Sessions from UI

fayimora:
- Create endpoint for creating events
- Create reference endpoints for Event subtypes(FSB & Telephone Interview types)


app/controllers/EventsController.scala
```diff
@@ -34,13 +34,16 @@ import scala.util.Try
 
 object EventsController extends EventsController {
   val eventsService: EventsService = EventsService
+  val eventsRepo: EventsMongoRepository = repositories.eventsRepository
```
kjharkin: I would say that an EventsController should not know about EventsService and EventsMongoRepository. Please refactor any direct calls on EventsMongoRepository into EventsService.

app/repositories/events/EventsConfigRepository.scala
```diff
@@ -123,10 +138,20 @@ trait EventsConfigRepository {
       )
       eventItemFuture.recoverWith {
```
kjharkin: Although this is always throwing an exception (if PartialFunction matches of course), I would argue that this should be ```recover``` not ```recoverWith``` - lets discuss

kjharkin:
Wii merge after comments addressed.

fayimora:
All done!

PR #431, opened 2017-07-26
FSET-1667: Remove candidates from sessions

fayimora:
Unfortunately, this PR got bigger than I would have liked but it was necessary as I took the liberty of decoupling assessor and candidate allocations(both services and repositories) and testing them in isolation properly. I also made some changes to the way candidates are added to a session base don some new progress states. 

Frontend PR: https://github.tools.tax.service.gov.uk/HMRC/fset-faststream-admin-frontend/pull/225


kjharkin:
LGTM :thumbsup:

PR #437, opened 2017-07-27
FSET-1634 Candidate Sift Forms

andrewrembrandt:


app/model/Exceptions.scala
```diff
@@ -108,6 +108,13 @@ object Exceptions {
 
   case class TooManySessionIdsException(m: String) extends Exception(m)
 
+  case class SchemeSpecificAnswerNotFound(m: String) extends Exception(m)
```
vglushak: frontend classes with the same names do not have `m` parameter. Is that expected?

app/model/Scheme.scala
```diff
@@ -19,7 +19,9 @@ package model
 import play.api.libs.json._
 import reactivemongo.bson.{ BSON, BSONHandler, BSONString }
 
-case class SchemeId(value: String)
+case class SchemeId(value: String) {
+  implicit override def toString: String = value
```
vglushak: I've never seen `implicit` for `toString`, could you please clarify why it's needed?

app/model/exchange/sift/SiftAnswers.scala
```diff
@@ -0,0 +1,52 @@
...
+
+package model.exchange.sift
+
+import model.exchange.sift.SiftAnswersStatus.SiftAnswersStatus
+import play.api.libs.json._
+
+object SiftAnswersStatus extends Enumeration {
+  type SiftAnswersStatus = Value
+
+  val DRAFT, SUBMITTED = Value
```
vglushak: Two similar enums. I've seen others put enum `AllocationStatuses` to `model` package, to avoid duplication. Don't know what is the best practice here.

andrewrembrandt: Am not comfortable using a type which is related to Allocation - this is purely a saved/committed document status - even so they might be representing the same values.

app/model/Scheme.scala
andrewrembrandt: to not wrap SchemeId(...) when serialising the value - it's used in the BSON form

app/model/Exceptions.scala
andrewrembrandt: Yeah - that text is wrapped into a 5XX/4XX response body, which should never be displayed on the frontend, but rather handled just by exception type

app/model/exchange/sift/SiftAnswers.scala
vglushak: I was talking about model.persisted.sift.SiftAnswersStatus and  model.exchange.sift.SiftAnswersStatus. AllocationStatuses was just like a sample.

andrewrembrandt: Discussed in person - leaving as it is for now, and will remove duplicated exchange/persisted model classes if they don't change in the future

app/repositories/CumulativeEvaluationHelper.scala
```diff
@@ -17,10 +17,10 @@
 package repositories
 
 import model.persisted.SchemeEvaluationResult
-import reactivemongo.bson.{ BSONArray, BSONDocument, BSONString }
+import reactivemongo.bson.{ BSONDocument, BSONString }
 
 trait CumulativeEvaluationHelper {
-  def cumulativeResultsForLatestPhaseBSON(latestResults: List[SchemeEvaluationResult]) = {
+  def cumulativeResultsForLatestPhaseBSON(latestResults: List[SchemeEvaluationResult]): BSONDocument = {
```
kjharkin: Need to change the name of this trait - as agreed before.

conf/app.routes
```diff
@@ -154,6 +154,15 @@ GET         /assessment-scores/with-candidate-summary/event/:eventId
 GET         /assessment-scores/:applicationId                                        controllers.AssessmentScoresController.find(applicationId: model.UniqueIdentifier)
 GET         /assessment-scores                                                       controllers.AssessmentScoresController.findAll
 
+# Scheme Specific Questions/Answers
+GET         /sift-answers/:applicationId/general                         controllers.SchemeSiftAnswersController.getGeneralAnswers(applicationId: String)
+POST        /sift-answers/:applicationId/general                         controllers.SchemeSiftAnswersController.addOrUpdateGeneralAnswers(applicationId: String)
```
kjharkin: Should be a PUT

conf/app.routes
```diff
@@ -154,6 +154,15 @@ GET         /assessment-scores/with-candidate-summary/event/:eventId
...
+GET         /sift-answers/:applicationId/status                          controllers.SchemeSiftAnswersController.getSiftAnswersStatus(applicationId: String)
+PUT         /sift-answers/:applicationId/submit                          controllers.SchemeSiftAnswersController.submitAnswers(applicationId: String)
+GET         /sift-answers/:applicationId/:schemeId                       controllers.SchemeSiftAnswersController.getSchemeSpecificAnswer(applicationId: String, schemeId: model.SchemeId)
+POST        /sift-answers/:applicationId/:schemeId                       controllers.SchemeSiftAnswersController.addOrUpdateSchemeSpecificAnswer(applicationId: String, schemeId: model.SchemeId)
```
kjharkin: Should be a PUT

app/repositories/CumulativeEvaluationHelper.scala
andrewrembrandt: That's on another branch

conf/app.routes
andrewrembrandt: Done

conf/app.routes
andrewrembrandt: Done

PR #438, opened 2017-07-27
FSET-1695 Update scheme representation with more attributes for front end

benjstephenson:


andrewrembrandt:
Discussed with Ben in person, all looks good.

PR #458, opened 2017-08-11
FSET-1666: Show & mark candidates that have been added to a specific FSB

WarFox:


conf/fsb-types.yaml
```diff
@@ -1,12 +1,39 @@
 - key: SAC
+  schemeId: GovernmentStatisticalService
```
vglushak: This will potentially lead to unsync. We have to manage schemes in this list and then in schemes yams file. 

I think it should go as-is, but in my story, I'll remove this file and use extended schemes structure:
https://github.com/hmrc/fset-faststream/commit/5a50eb034b3a789761094d57e49257d41e0dd718#diff-1e6cb5c6d8ee215686f6a13e5a2b99b6
If there are any objections, please let me know.

fayimora: @vglushak How far into your story are you? Do you think you can PR that change separately so it's integrated into this branch?

it/repositories/application/FsbTestGroupMongoRepositorySpec.scala
```diff
@@ -0,0 +1,142 @@
...
+      val Some(result) = repository.findByApplicationId(applicationId).futureValue
+      val expectedFsbTestGroup = FsbTestGroup(List(schemeEvaluationResult1, schemeEvaluationResult2))
+      result mustBe expectedFsbTestGroup
+    }
+
+    "not overwrite existing value" in {
+      val applicationId = createApplication()
+      repository.save(applicationId, SchemeEvaluationResult("GovernmentOperationalResearchService", "Green")).futureValue
+
+      intercept[Exception] {
```
fayimora: Because you are throwing a generic exception here, this test is not realllllly valid. You'll need to throw a custom exception and intercept that.

conf/fsb-types.yaml
WarFox: good idea 👍

it/repositories/application/FsbTestGroupMongoRepositorySpec.scala
```diff
@@ -0,0 +1,142 @@
...
+
+      result must contain theSameElementsAs expectedResult
+    }
+
+  }
+
+  private def applicationRepository = new GeneralApplicationMongoRepository(DateTimeFactory, cubiksGatewayConfig)
+
+  private def createApplication(): String = {
+    val applicationId = generateUUID()
```
fayimora: Minor thing and you probably don't have to change but I prefer to invoke `UUIDFactory.generateUUID()` as opposed to mixing it in.

test/controllers/FsbTestGroupControllerSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+  val mockEventsService = mock[EventsService]
+
+  val controller = new FsbTestGroupController {
+    val eventsService = mockEventsService
+    val service = mockFsbTestGroupService
+  }
+
+  "save fsb event evaluation result" should {
+    "return Created when save is successful" in {
+    }
```
fayimora: Empty test

app/controllers/FsbTestGroupController.scala
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  def find(applicationIds: List[String], fsbType: Option[String]): Action[AnyContent] = Action.async { implicit request =>
+    service.findByApplicationIdsAndFsbType(applicationIds, fsbType).map { results =>
+      Ok(Json.toJson(results))
+    }
+  }
+
+  private def fromPassMark(s: String): EvaluationResults.Result = s match {
```
WarFox: The Same thing is here... 
https://github.com/hmrc/fset-faststream/blob/master/app/controllers/SiftingController.scala#L51

I agree we should make it somewhere central, may be in [model.EvaluationResult.scala](https://github.com/hmrc/fset-faststream/blob/master/app/model/EvaluationResults.scala)

it/repositories/application/FsbTestGroupMongoRepositorySpec.scala
WarFox: We are throwing a custom exception here... https://github.com/hmrc/fset-faststream/pull/458/files#diff-fb2ebf497e9c888ce19dc932e62d089eR57

I forgot to update test, will update test with custom exception

app/controllers/FsbTestGroupController.scala
```diff
@@ -0,0 +1,66 @@
...
+    case "Pass" => EvaluationResults.Green
+    case "Fail" => EvaluationResults.Red
```
WarFox: sure!

app/controllers/FsbTestGroupController.scala
```diff
@@ -0,0 +1,66 @@
...
+    case _ => sys.error(s"Unsupported evaluation result $s for FSB")
```
WarFox: Will go away, when we use Enum

app/services/application/FsbTestGroupService.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  def findByApplicationIdsAndFsbType(applicationIds: List[String], fsbType: Option[String]): Future[List[FsbSchemeResult]] = {
+    val eventualMayBeSchemeId = for {
+      fsbTypes <- eventsService.getFsbTypes
+      fsb <- Future(fsbTypes.find(f => fsbType.contains(f.key)))
+      schemeId <- Future(fsb.map(f => SchemeId(f.schemeId)))
+    } yield schemeId
+
+    eventualMayBeSchemeId.flatMap { mayBeSchemeId =>
+      findByApplicationIdsAndScheme(applicationIds, mayBeSchemeId)
```
WarFox: yes 👍

fayimora:
I'll merge once all ATs pass

app/services/application/FsbTestGroupService.scala
```diff
@@ -0,0 +1,63 @@
...
+object FsbTestGroupService extends FsbTestGroupService {
+  override val fsbTestGroupRepository = repositories.fsbTestGroupRepository
+  override val eventsService = EventsService
+}
+
+trait FsbTestGroupService {
+  val fsbTestGroupRepository: FsbTestGroupRepository
+  val eventsService: EventsService
+
+  def saveResults(schemeId: SchemeId, applicationResults: List[ApplicationResult]): Future[Unit] = {
```
henricook: This is orphaning futures. All futures must be mapped somewhere - this method should either Future.sequence and map over the saveResults calls, or return a Future[List[Unit]] (wouldn't be my preferred, i'd use the sequence)

app/services/application/FsbTestGroupService.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  def saveResult(schemeId: SchemeId, applicationResult: ApplicationResult): Future[Unit] = {
+    val schemeEvaluationResult = SchemeEvaluationResult(schemeId, applicationResult.result)
+    fsbTestGroupRepository.save(applicationResult.applicationId, schemeEvaluationResult)
+  }
+
+  def findByApplicationIdsAndFsbType(applicationIds: List[String], fsbType: Option[String]): Future[List[FsbSchemeResult]] = {
+    for {
+      fsbTypes <- eventsService.getFsbTypes
+      fsb <- Future(fsbTypes.find(f => fsbType.contains(f.key)))
```
henricook: Why not combine these two lines?

app/services/events/EventsService.scala
```diff
@@ -95,4 +97,15 @@ trait EventsService {
   def getFsbTypes: Future[List[FsbType]] = eventsConfigRepo.fsbTypes
 
   def getTelephoneInterviewTypes: Future[List[TelephoneInterviewType]] = eventsConfigRepo.telephoneInterviewTypes
+
+  def findSchemeByEvent(eventId: String): Future[Option[Scheme]] = {
+    for {
+      event <- getEvent(eventId)
+      fsbTypes <- getFsbTypes
+      maybeFsbType <- Future(fsbTypes.find(f => f.key == event.description))
+      fsbType: FsbType = maybeFsbType.getOrElse(throw new Exception(s"FsbType with event description ${event.description} not found"))
```
henricook: Can you combine these two lines (maybeFsbType and fsbType)? I think if it goes wrong you'll end in a failed future, with the exception inside

app/controllers/FsbTestGroupController.scala
```diff
@@ -0,0 +1,61 @@
...
+      val greenRedResults = fsbEvaluationResults.applicationResults.map { applicationResult =>
+        applicationResult.copy(result = EvaluationResults.Result.fromPassFail(applicationResult.result).toString)
+      }
+
+      eventsService.findSchemeByEvent(eventId).flatMap {
+        case Some(scheme) => service.saveResults(scheme.id, greenRedResults)
+        case None => throw new SchemeNotFoundException(s"Event $eventId has no associated Scheme. FsbType mismatch")
+      }
+
+      Future.successful(Ok)
```
kjharkin: No - this is incorrect

henricook: Ah indeed! It's another orphaned future right? So we should be mapping the results of the findSchemeByEvent flatMap to Ok?

kjharkin: yep

app/services/application/FsbTestGroupService.scala
kjharkin: Indeed - you are creating Futures artificially here. Something like 
```
def findByApplicationIdsAndFsbType(applicationIds: List[String], fsbType: Option[String]): Future[List[FsbSchemeResult]] = {
    for {
      fsbTypes <- eventsService.getFsbTypes
      fsb <- fsbTypes.find(f => fsbType.contains(f.key))
      fsbSchemes <- findByApplicationIdsAndScheme(applicationIds, Some(SchemeId(fsb.schemeId)))
    } yield fsbSchemes
  }
```

henricook: This will mix future and non future though i think - will it bork at that?

What's wrong with `fsbTypes <- eventsService.getFsbTypes.map(_.find(f => fsbType.contains(f.key)))` thinking about it? I think splitting it out into steps is adding complexity even if it might slightly improve readability

app/controllers/FsbTestGroupController.scala
WarFox: this was something I was supposed to refactor but missed it. Sorry about that!

app/services/events/EventsService.scala
WarFox: I'll also remove the Future wrapping there, like in controller

WarFox: Is this readable?

```
  def findSchemeByEvent(eventId: String): Future[Option[Scheme]] = {
    for {
      event <- getEvent(eventId)
      fsbTypes <- getFsbTypes
    } yield {
      fsbTypes.collectFirst {
        case fsbType if fsbType.key == event.description => SchemeId(fsbType.schemeId)
      } match {
        case Some(schemeId) => schemeRepo.getSchemeForId(schemeId)
        case _ => throw new Exception(s"FsbType with event description ${event.description} not found")
      }
    }
  }```

File: app/services/application/FsbTestGroupService.scala (hunk 10 above)
Comment (WarFox) [inline app/services/application/FsbTestGroupService.scala]:
```scala
    eventsService.getFsbTypes.flatMap { fsbTypes =>
      val mayBeSchemeId = fsbTypes.collectFirst { case x if mayBeFsbType.contains(x.key) => SchemeId(x.schemeId) }
      findByApplicationIdsAndScheme(applicationIds, mayBeSchemeId)
    }
```

Rewriting it like this!

app/controllers/FsbTestGroupController.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  def save(eventId: String, sessionId: String): Action[JsValue] = Action.async(parse.json) { implicit request =>
+    withJsonBody[FsbEvaluationResults] { fsbEvaluationResults =>
+      val greenRedResults = fsbEvaluationResults.applicationResults.map { applicationResult =>
+        applicationResult.copy(result = EvaluationResults.Result.fromPassFail(applicationResult.result).toString)
+      }
+
+      eventsService.findSchemeByEvent(eventId).flatMap {
+        case Some(scheme) => fsbService.saveResults(scheme.id, greenRedResults)
+        case None => throw new SchemeNotFoundException(s"Event $eventId has no associated Scheme - FsbType mismatch")
```
fayimora: I don't think you need the `new` modifier since the exception is a case class

app/controllers/FsbTestGroupController.scala
```diff
@@ -0,0 +1,63 @@
...
+  val eventsService: EventsService
+
+  def save(eventId: String, sessionId: String): Action[JsValue] = Action.async(parse.json) { implicit request =>
+    withJsonBody[FsbEvaluationResults] { fsbEvaluationResults =>
+      val greenRedResults = fsbEvaluationResults.applicationResults.map { applicationResult =>
+        applicationResult.copy(result = EvaluationResults.Result.fromPassFail(applicationResult.result).toString)
+      }
+
+      eventsService.findSchemeByEvent(eventId).flatMap {
+        case Some(scheme) => fsbService.saveResults(scheme.id, greenRedResults)
```
fayimora: Isn't this future orphaned too? cc @henricook

PR #467, opened 2017-08-16
Part of FSET-1689: Set progress status of the candidate when fsb results are entered

henricook:


henricook:
@kjharkin ready to merge

PR #471, opened 2017-08-17
FSET-1658. Show only candidates with first residual preference the sa…

vglushak:
…me as event sub type.

### DIFF
--- a/app/repositories/CurrentSchemeStatusHelper.scala
+++ b/app/repositories/CurrentSchemeStatusHelper.scala
@@ -66,4 +66,16 @@ trait CurrentSchemeStatusHelper {
       doc ++ BSONDocument(s"currentSchemeStatus" -> BSONDocument("$elemMatch" -> SchemeEvaluationResult(id, status.toString)))
     }
   }
+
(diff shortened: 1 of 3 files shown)


app/repositories/CurrentSchemeStatusHelper.scala
```diff
@@ -66,4 +66,16 @@ trait CurrentSchemeStatusHelper {
       doc ++ BSONDocument(s"currentSchemeStatus" -> BSONDocument("$elemMatch" -> SchemeEvaluationResult(id, status.toString)))
     }
   }
+
+  def firstResidualPreference(schemeId: SchemeId): BSONDocument = {
+    BSONDocument("$where" ->
+      s"""
+        |this.currentSchemeStatus.filter(
+        |   function(e){
+        |     return e.result=="$Green"
```
benjstephenson: What's this $ for on the Green?

vglushak: I refer to `Green` object which is defined somewhere in the system.

PR #474, opened 2017-08-18
FSET-1754: Return assessor allocation report with flexible columns

henricook:


app/controllers/ReportingController.scala
```diff
@@ -87,6 +95,59 @@ trait ReportingController extends BaseController {
     }
   }
 
+  private def makeRow(values: Option[String]*) =
+    values.map { s =>
+      val ret = s.getOrElse(" ").replace("\r", " ").replace("\n", " ").replace("\"", "'")
```
benjstephenson: You could probably collapse the separate replace calls with one with regex

henricook: TBH i just copied this from the admin frontend, will look

app/controllers/ReportingController.scala
```diff
@@ -87,6 +95,59 @@ trait ReportingController extends BaseController {
...
+
+  def assessorAllocationReport: Action[AnyContent] = Action.async { implicit request =>
+
+    import common.Joda._
+
+    val sortedEventsFut = eventsRepository.findAll.map(_.sortBy(_.date))
+
+    val reportRows = for {
+      allAssessors <- assessorRepository.findAll
+      allAssessorsPersonalInfo <- authProviderClient.findByUserIds(allAssessors.map(_.userId)).map(_.groupBy(_.userId).mapValues(_.head))
```
benjstephenson: Why do we need to group by the userId?  Surely there'll only be one response per userId?

henricook: Yes but i wanted a Map for more efficient lookups

app/controllers/ReportingController.scala
henricook: I'm not sure there's a particularly more efficient way to do this - feel free to recommend something, i could combine \r and \n in a regex but i'm not sure it's a substantial gain

app/controllers/ReportingController.scala
benjstephenson: you could reduce the ```groupBy(_.userId).mapValues(_.head)``` bit to ```map(x => x.userId -> x)(breakOut)``` to avoid the extra collection traversals, I think

app/repositories/AssessorAllocationRepository.scala
```diff
@@ -62,6 +63,11 @@ class AssessorAllocationMongoRepository(implicit mongo: () => DB)
     collection.find(query, projection).one[AssessorAllocation]
   }
 
+  def findAll: Future[List[AssessorAllocation]] = {
```
benjstephenson: You should get this method for free because we extend ReactiveRepository

app/repositories/AssessorRepository.scala
```diff
@@ -51,6 +52,12 @@ class AssessorMongoRepository(implicit mongo: () => DB)
     collection.find(query).one[Assessor]
   }
 
+  def findAll: Future[List[Assessor]] = {
```
benjstephenson: Ditto here

app/repositories/events/EventsRepository.scala
```diff
@@ -48,6 +49,12 @@ class EventsMongoRepository(implicit mongo: () => DB)
       .map(_ => ())
   }
 
+  def findAll: Future[List[Event]] = {
```
benjstephenson: Hat-trick!

app/repositories/AssessorAllocationRepository.scala
henricook: Amended

app/repositories/AssessorRepository.scala
henricook: Amended

app/controllers/ReportingController.scala
henricook: Done - i do like efficiency

henricook:
Review changes added

app/controllers/ReportingController.scala
fayimora: @henricook @benjstephenson 
Didn't this do the trick?

```scala
  input.replace("\r|\n|\\"", " ")
```

PR #475, opened 2017-08-18
FSET-1658. Handle case when candidate is eligible for FSB but there a…

vglushak:
…re no Green scheme.


vglushak:
@kjharkin - you should probably see it.

PR #479, opened 2017-08-18
FSET-1738 - Added clearCandidates to test data generator

andrewrembrandt:


fayimora:
This clears only the application but not the corresponding user from auth provider.

app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -907,6 +910,12 @@ class GeneralApplicationMongoRepository(
     replaceAllocationStatus(applicationId, EventProgressStatuses.get(eventType.applicationStatus).failedToAttend)
   }
 
+  override def remove(applicationStatus: Option[String]): Future[Int] = {
+    val query = applicationStatus.map(as => BSONDocument("applicationStatus" -> as)).getOrElse(BSONDocument())
```
fayimora: I'll prefer to call this `appStatus` or `status`.

andrewrembrandt:
I've added the call to Auth Provider to delete - just need to test quickly once I'm in the office

henricook:
What's your use case for this?

henricook:
Can you plaster a warning comment on both the authproviderclient method and generalapplicationrepository method saying this is for test data reasons only, we don't want anyone ever using these methods by accident in features

henricook:
I notice the DELETE endpoint in the auth provider isn't in test only routes which would make me feel beter - it was added by @mgilgar  - Miguel do we need that endpoint to be a production endpoint for some feature?

andrewrembrandt:
DELETE was added by @mgilgar when he did the CreateAssessor work in June - it's used in the admin F/E when user creation failed on the backend (i.e. it removes the associated user), so is required in production.

All tests pass, @kjharkin, ready for merge (with the admin AT PR also)

app/repositories/application/GeneralApplicationRepository.scala
andrewrembrandt: So, that in fact totally opposite of what we should be doing - the Robert C Martin, books, LSCC, and top consultants advocate use of full naming - i.e. no ambiguity

henricook: I'm going to come out on the side of full names

henricook:
Something about having a removeAllUsersInStatus method available to be used makes me really nervous

henricook:
Is it worth adding a feature flag to ensure this can *never* be used accidentally in production @kjharkin , or am i overreacting?

kjharkin:
Can we move this remove to a specific Test.... repository please.

andrewrembrandt:
@kjharkin, can't put it into 'test' or 'it' scope as it's used via TestDataGenerator, so I've moved 'remove' into a dedicated repo that exists in a 'testdata' package which matches with the TDG package.

henricook:
:muscle: - you had a conflict though

PR #481, opened 2017-08-21
FSET-1744 Progress To FSAC after withdrawal

benjstephenson:
When withdrawing from schemes during sift, if there are only HR or Generalist left in the scheme list then progress the candidate to FSAC allocation.


kjharkin:
LGTM :thumbsup:

PR #485, opened 2017-08-22
NoJira: Multiple set statements do not work in mongo, so rework the sift update

benjstephenson:
BSON to ensure all sub-docs are correctly updated.

### DIFF
--- a/app/services/sift/ApplicationSiftService.scala
+++ b/app/services/sift/ApplicationSiftService.scala
@@ -94,7 +94,6 @@ trait ApplicationSiftService extends CurrentSchemeStatusHelper with CommonBSONDo
         maybeSetProgressStatus(siftedSchemes.toSet, candidatesSiftableSchemes.toSet)
       )
 
-      play.api.Logger.error(s"\n\nBLAH : ${settableFields.map(BSONDocument.pretty)}")
       applicationSiftRepo.siftApplicationForScheme(applicationId, result, settableFields)
 
     }) flatMap identity
--- a/app/services/sift/ApplicationSiftService.scala
+++ b/app/services/sift/ApplicationSiftService.scala
@@ -76,7 +76,7 @@ trait ApplicationSiftService extends CurrentSchemeStatusHelper with CommonBSONDo
 
     def maybeSetProgressStatus(siftedSchemes: Set[SchemeId], siftableSchemes: Set[SchemeId]) =
       if (siftedSchemes equals  siftableSchemes) {
-        BSONDocument("$set" -> progressStatusOnlyBSON(ProgressStatuses.SIFT_COMPLETED))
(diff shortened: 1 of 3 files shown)


app/services/sift/ApplicationSiftService.scala
```diff
@@ -76,28 +76,27 @@ trait ApplicationSiftService extends CurrentSchemeStatusHelper with CommonBSONDo
...
       val siftedSchemes = (currentSiftEvaluation.map(_.schemeId) :+ result.schemeId).distinct
 
-      val mergedUpdate = Seq(
-        BSONDocument("$set" -> currentSchemeStatusBSON(newSchemeStatus)),
+      val settableFields = Seq(currentSchemeStatusBSON(newSchemeStatus),
         maybeSetProgressStatus(siftedSchemes.toSet, candidatesSiftableSchemes.toSet)
-      ).foldLeft(siftBson) { (acc, doc) => acc ++ doc }
+      )
+
+      play.api.Logger.error(s"\n\nBLAH : ${settableFields.map(BSONDocument.pretty)}")
```
vglushak: BLAH? is that 'Bringing Life And Hope' or 'Boring Links and Homepages'? :)

benjstephenson: Oops, I'll remove that!

vglushak:
I'll merge as soon as checks passed.

test/services/sift/ApplicationSiftServiceSpec.scala
```diff
@@ -65,11 +65,8 @@ class ApplicationSiftServiceSpec extends ScalaMockUnitSpec {
 
   trait SiftUpdateTest extends TestFixture {
     val progressStatusUpdateBson = BSONDocument(
-      "$set" -> BSONDocument(
-        s"progress-status.${ProgressStatuses.SIFT_COMPLETED}" -> true,
-        s"progress-status-timestamp.${ProgressStatuses.SIFT_COMPLETED}" ->
-          BSONDateTimeHandler.write(DateTimeFactoryMock.nowLocalTimeZone)
-      )
+      s"progress-status.${ProgressStatuses.SIFT_COMPLETED}" -> true,
```
henricook: Is this correct? It looks like it's no longer doing a set and is doing an overwrite

henricook: Technically i think we're meant to be doing  { progress-status: { ProgressStatuses.SIFT_COMPLETED: true } } or the keys set all wonky, but i think we're doing that wrong *everywhere*

benjstephenson: Yup, the repo method that this service calls will wrap all the fields that are to be set in a single $set operator

PR #496, opened 2017-08-23
FSET-1773 Scheduled job for candidates who have failed all schemes at SIFT_COMPLETE

benjstephenson:


app/services/sift/ApplicationSiftService.scala
```diff
@@ -34,21 +36,28 @@ import scala.language.postfixOps
...
   def schemeRepo: SchemeRepository
+  def emailClient: CSREmailClient
 
   def nextApplicationsReadyForSiftStage(batchSize: Int): Future[Seq[ApplicationForSift]] = {
     applicationSiftRepo.nextApplicationsForSiftStage(batchSize)
   }
 
+  def processNextApplicationFailedAtSift: Future[Unit] = applicationSiftRepo.nextApplicationFailedAtSift.flatMap(_.map { application =>
+    applicationRepo.addProgressStatusAndUpdateAppStatus(application.applicationId, ProgressStatuses.FAILED_AT_SIFT)
+  }.getOrElse(Future()))
```
vglushak: should it be `Future.success` ?

app/model/ProgressStatuses.scala
```diff
@@ -111,6 +111,8 @@ object ProgressStatuses {
   case object SIFT_ENTERED extends ProgressStatus(ApplicationStatus.SIFT)
   case object SIFT_READY extends ProgressStatus(ApplicationStatus.SIFT)
   case object SIFT_COMPLETED extends ProgressStatus(ApplicationStatus.SIFT)
+  case object FAILED_AT_SIFT extends ProgressStatus(ApplicationStatus.FAILED_AT_SIFT)
+  case object SDIP_FAILED_AT_SIFT extends ProgressStatus(ApplicationStatus.SIFT)
```
vglushak: I've seen there a couple more examples of adding scheme information to Progress status, but is it essential
here to have sdip specific progress status? Is that for reports?

benjstephenson: We need it for Sdip as it's not just an additional scheme, it's a whole application in itself.

app/services/sift/ApplicationSiftService.scala
benjstephenson: This will be a Future success as there is no exception, just a unit value.

app/model/ProgressStatuses.scala
```diff
@@ -111,6 +111,8 @@ object ProgressStatuses {
   case object SIFT_ENTERED extends ProgressStatus(ApplicationStatus.SIFT)
   case object SIFT_READY extends ProgressStatus(ApplicationStatus.SIFT)
   case object SIFT_COMPLETED extends ProgressStatus(ApplicationStatus.SIFT)
+  case object FAILED_AT_SIFT extends ProgressStatus(ApplicationStatus.FAILED_AT_SIFT)
```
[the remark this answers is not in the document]
benjstephenson: The difference here is that conceptually you can't fail a sift as there are no pass marks.  Failed implies to me that there was some issue doing the sift, not that all of their schemes were red once sift was complete.

app/services/sift/ApplicationSiftService.scala
vglushak: Future.success returns already completed future, while Future() - create another empty thread and execute it. Few extra cpu cycles, Do we need them?

vglushak:
I'll merge as soon as all tests pass.

app/services/sift/ApplicationSiftService.scala
benjstephenson: 👍  TIL.  Will change

PR #498, opened 2017-08-24
Updated scheme names (and remove International)

charge-valtech:
This removes International as a scheme, updates DaT to Digital, Data & Technology, and gives GES_DS its proper name of Diplomatic Service (Economists). Also updated year for SDIP. This PR has corresponding frontend changes.


vglushak:
fixing tests for that.

PR #501, opened 2017-08-24
FSET-1644 Corrected sift progression to SIFT_COMPLETE - factor in sift fails along with passes

andrewrembrandt:
…ong with passes


WarFox:
Looks good to me... waiting for tests to run

PR #502, opened 2017-08-24
FSET-1764. Send reminder 5 days after event invite

vglushak:


app/model/persisted/Allocation.scala
```diff
@@ -56,7 +57,9 @@ case class CandidateAllocation(
   sessionId: String,
   status: AllocationStatus,
   version: String,
-  removeReason: Option[String]
+  removeReason: Option[String],
+  createdAt: LocalDate,
+  isNotified: Boolean
```
kjharkin: ```isNotified``` isn't intention revealing. Can you rename this to something else such as ```reminderSent```?

app/services/allocation/CandidateAllocationService.scala
```diff
@@ -197,6 +202,23 @@ trait CandidateAllocationService extends EventSink {
     )
   }
 
+
+  def processUnconfirmedCandidates()(implicit hc: HeaderCarrier, rh: RequestHeader): Future[Unit] = {
+    candidateAllocationRepo.findAllUnconfirmedAllocated(eventsConfig.daysBeforeInvitationReminder).flatMap { allocations =>
+      val rr = Future.sequence(allocations.map { alloc =>
+        val r = eventsService.getEvent(alloc.eventId).flatMap { event =>
```
kjharkin: we don't need the ```r``` and ```rr``` do we?

vglushak:
comments addressed.

PR #504, opened 2017-08-25
FSET-1711: Progress to FSB or Job offer from assessment_centre_passed

henricook:


app/repositories/application/GeneralApplicationRepository.scala
```diff
@@ -225,21 +225,25 @@ class GeneralApplicationMongoRepository(
         val applicationStatus = document.getAs[ApplicationStatus]("applicationStatus").get
         val applicationRoute = document.getAs[ApplicationRoute]("applicationRoute").getOrElse(ApplicationRoute.Faststream)
         val progressStatusTimeStampDoc = document.getAs[BSONDocument]("progress-status-timestamp")
-        val progressStatusTimeStamp = progressStatusTimeStampDoc.flatMap { timestamps =>
+        val latestProgressStatus = progressStatusTimeStampDoc.flatMap { timestamps =>
           val relevantProgressStatuses = timestamps.elements.filter(_._1.startsWith(applicationStatus))
+          val latestRelevantProgressStatus = relevantProgressStatuses.maxBy(element => timestamps.getAs[DateTime](element._1).get)
+          Try(ProgressStatuses.nameToProgressStatus(latestRelevantProgressStatus._1)) match {
+            case Success(progressStatus) => Some(progressStatus)
```
vglushak: You can use `.toOption` function on `Try()`

app/services/assessmentcentre/AssessmentCentreToFsbOrOfferProgressionService.scala
```diff
@@ -0,0 +1,71 @@
...
+    fsbRepo.nextApplicationForFsbOrJobOfferProgression(batchSize)
+  }
+
+  def progressApplicationsToFsbOrJobOffer(applications: Seq[ApplicationForProgression])
+  : Future[SerialUpdateResult[ApplicationForProgression]] = {
+
+    def maybeProgressToFsbOrJobOffer(application: ApplicationForProgression, firstResidual: SchemeEvaluationResult): Future[Unit] = {
+      if (firstResidual.result == Green.toString && fsbRequiredSchemeIds.contains(firstResidual.schemeId)) {
+        fsbRepo.progressToFsb(application)
+      } else if (firstResidual.result == Green.toString && !fsbRequiredSchemeIds.contains(firstResidual.schemeId)) {
```
vglushak: second condition looks excessive.
I would do it as pattern matching on `result` with guard condition on `fsbRequiredSchemeIds` check.

app/services/assessmentcentre/AssessmentCentreToFsbOrOfferProgressionService.scala
```diff
@@ -0,0 +1,71 @@
...
+      } else {
+        Future.successful(())
+      }
+    }
+
+    val updates = FutureEx.traverseSerial(applications) { application =>
+      FutureEx.futureToEither(application,
+        for {
+          currentSchemeStatus <- applicationRepo.getCurrentSchemeStatus(application.applicationId)
+          firstResidual = firstResidualPreference(currentSchemeStatus).get
```
vglushak: what if candidate has failed everything - `None.get` ?

app/model/ProgressStatuses.scala
```diff
@@ -121,6 +121,7 @@ object ProgressStatuses {
   case object ASSESSMENT_CENTRE_AWAITING_RE_EVALUATION extends ProgressStatus(ApplicationStatus.ASSESSMENT_CENTRE)
   case object ASSESSMENT_CENTRE_PASSED extends ProgressStatus(ApplicationStatus.ASSESSMENT_CENTRE)
   case object ASSESSMENT_CENTRE_FAILED extends ProgressStatus(ApplicationStatus.ASSESSMENT_CENTRE)
+  case object ASSESSMENT_CENTRE_FAILED_NOTIFIED extends ProgressStatus(ApplicationStatus.ASSESSMENT_CENTRE)
```
vglushak: Thanks for adding this for me.

I know don't look a gift-horse in the mouth, but maybe you can also add `FSB_FAILED_NOTIFIED` 😳

app/repositories/application/GeneralApplicationRepository.scala
henricook: :+1:

app/services/assessmentcentre/AssessmentCentreToFsbOrOfferProgressionService.scala
henricook: Shouldn't be possible at this stage, they shouldn't be getting passed to this function without an eligible scheme. In that case this .get fails and it comes back as a failure in SerialUpdateResult

app/model/ProgressStatuses.scala
henricook: Go on then :roll_eyes:

PR #506, opened 2017-08-29
FSET-1731: Schemes without evaluation should be progressed to the right state

fayimora:
Some schemes do not require an evaluation(no pass/fail). If these schemes
are the green schemes after sift or form submission, the candidate should be
progressed to the right state.

### DIFF
--- a/app/services/sift/SiftAnswersService.scala
+++ b/app/services/sift/SiftAnswersService.scala
@@ -75,10 +75,25 @@ trait SiftAnswersService {
       currentSchemeStatus <- appRepo.getCurrentSchemeStatus(applicationId)
       schemesPassed = currentSchemeStatus.filter(_.result == EvaluationResults.Green.toString).map(_.schemeId).toSet
       schemesPassedRequiringSift = schemeRepository.schemes.filter( s =>
-        schemesPassed.contains(s.id) && s.siftRequirement == Some(SiftRequirement.FORM)
+        schemesPassed.contains(s.id) && s.siftRequirement.contains(SiftRequirement.FORM)
+      ).map(_.id).toSet
+      schemesPassedWithoutRequirement = schemeRepository.schemes.filter( s =>
+        schemesPassed.contains(s.id) && !s.siftEvaluationRequired
       ).map(_.id).toSet
       _ <- siftAnswersRepo.submitAnswers(applicationId, schemesPassedRequiringSift)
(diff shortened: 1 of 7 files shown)


henricook:
Conflicts @fayimora

fayimora:
@henricook Fixed!

app/services/sift/SiftAnswersService.scala
```diff
@@ -75,10 +75,25 @@ trait SiftAnswersService {
...
   }
+
+  private def maybeMoveToCompleted(applicationId: String, passedSchemes: Set[SchemeId],
+                                   passedSchemesWithoutRequirement: Set[SchemeId]): Future[Unit] = {
+    val canBeMoved = (passedSchemes.size == passedSchemesWithoutRequirement.size) &&
+      passedSchemes.forall(passedSchemesWithoutRequirement.contains)
+    if(canBeMoved) {
+      appRepo.addProgressStatusAndUpdateAppStatus(applicationId, ProgressStatuses.SIFT_COMPLETED)
+    } else {
+      Future(())
```
benjstephenson: return future success explicitly or you'll create another thread.

app/services/sift/SiftAnswersService.scala
```diff
@@ -75,10 +75,25 @@ trait SiftAnswersService {
...
       ).map(_.id).toSet
       _ <- siftAnswersRepo.submitAnswers(applicationId, schemesPassedRequiringSift)
       _ <- appRepo.addProgressStatusAndUpdateAppStatus(applicationId, ProgressStatuses.SIFT_READY)
+      _ <- maybeMoveToCompleted(applicationId, schemesPassed, schemesPassedWithoutRequirement)
     } yield {}
   }
+
+  private def maybeMoveToCompleted(applicationId: String, passedSchemes: Set[SchemeId],
+                                   passedSchemesWithoutRequirement: Set[SchemeId]): Future[Unit] = {
+    val canBeMoved = (passedSchemes.size == passedSchemesWithoutRequirement.size) &&
```
benjstephenson: belt and braces, but - we should really be checking that the sets are the same, not just the same size.

app/services/sift/SiftAnswersService.scala
```diff
@@ -75,10 +75,25 @@ trait SiftAnswersService {
...
+      passedSchemes.forall(passedSchemesWithoutRequirement.contains)
```
benjstephenson: Use set equality rather than iterating over them.

benjstephenson:
@fayimora a few comments - Can you also write some unit tests please.

PR #511, opened 2017-08-30
FSET-1746 Update FSAC indicator

ian-stainer:


app/services/personaldetails/PersonalDetailsService.scala
```diff
@@ -98,4 +98,15 @@ trait PersonalDetailsService {
       _ <- civilServiceExperienceDetailsUpdateFut
     } yield {}
   }
+
+  def updateFsacIndicator(applicationId: String, userId: String, fsacArea: String): Future[Unit] = {
+    val validFsacAreas = Seq("London", "Newcastle")
```
benjstephenson: I think this would be better placed in config somewhere.

PR #515, opened 2017-08-31
FSET-1387. Notify candidate of FSAC results - 1st pref. needs FSB.

vglushak:


app/services/assessmentcentre/AssessmentCentreToFsbOrOfferProgressionService.scala
```diff
@@ -59,14 +77,27 @@ trait AssessmentCentreToFsbOrOfferProgressionService extends CurrentSchemeStatus
...
       )
     }
-
     updates.map(SerialUpdateResult.fromEither)
   }
-}
+
+  private def retrieveCandidateDetails(applicationId: String)(implicit hc: HeaderCarrier) = {
+    applicationRepo.find(applicationId).flatMap {
+      case Some(app) => authProviderClient.findByUserIds(Seq(app.userId))
```
benjstephenson: You can get these details from the application document, you don't need to add a dependency on the auth provider

vglushak: Actually I can't - there are no email there.

benjstephenson: candidate details collection

vglushak: ok, but that means I have to add dependency to contact-details repository, isn't it?

benjstephenson:
True, but means that we're not calling another service outside of FS
backend.

app/services/assessmentcentre/AssessmentCentreToFsbOrOfferProgressionService.scala
benjstephenson: True, but it means we don't have to call an external service

test/services/assessmentcentre/AssessmentCentreToFsbOrOfferProgressionServiceSpec.scala
```diff
@@ -38,10 +41,22 @@ class AssessmentCentreToFsbOrOfferProgressionServiceSpec extends ScalaMockUnitSp
...
+          .expects(expectedApplication.applicationId)
+          .returningAsync(Option(c0)).once
+
+        (mockAuthProviderClient.findByUserIds(_: Seq[String])(_: HeaderCarrier))
+          .expects(Seq(userId), hc)
+          .returningAsync(Seq(c1)).once
+
+        (mockEmailClient.sendCandidateAssessmentCompletedMovedToFsb(_: String, _: String)(_: HeaderCarrier))
+          .expects(c1.email, c1.name, hc)
+          .returningAsync.once
```
benjstephenson: I'm pretty sure that you can omit the .once calls and the default is that the method is only called once

vglushak:
Both comments addressed!

benjstephenson:
LGTM 👍  Will wait for Travis to be happy and then merge

PR #518, opened 2017-09-01
Fset 1776 - Add Current Scheme Status to Diversity Report

andrewrembrandt:


app/repositories/application/ReportingRepoBSONReader.scala
```diff
@@ -186,10 +186,13 @@ trait ReportingRepoBSONReader extends CommonBSONDocuments with BaseBSONReader {
       val userId = doc.getAs[String]("userId").getOrElse("")
       val progress: ProgressResponse = toProgressResponse(applicationId).read(doc)
 
+      val curSchemeStatus = doc.getAs[List[SchemeEvaluationResult]]("currentSchemeStatus").getOrElse(
+        schemes.getOrElse(List.empty).map(s => new SchemeEvaluationResult(s, EvaluationResults.Green.toString)))
```
benjstephenson: I don't think we should be faking results here.

PR #519, opened 2017-09-04
FSET-1662: Edit Event Capacity

fayimora:


app/model/persisted/eventschedules/Event.scala
```diff
@@ -39,7 +39,21 @@ case class Event(
   skillRequirements: Map[String, Int],
   sessions: List[Session],
   wasBulkUploaded: Boolean = false
-)
+) {
+  def session(sessionId: String): Session = {
```
andrewrembrandt: As discussed, please move elsewhere

PR #527, opened 2017-09-06
FSET-1791 Allow Assessors to add availability for wfh

benjstephenson:
Artificially add a day aggregated event for the "Home" location for each day that has another event running.

### DIFF
--- a/app/controllers/DayAggregateEventController.scala
+++ b/app/controllers/DayAggregateEventController.scala
@@ -16,7 +16,6 @@
 
 package controllers
 
-import config.MicroserviceAppConfig
 import model.persisted.eventschedules.Location
 import model.persisted.eventschedules.SkillType.SkillType
 import org.joda.time.LocalDate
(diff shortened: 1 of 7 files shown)


andrewrembrandt:
LGTM

app/controllers/DayAggregateEventController.scala
```diff
@@ -47,8 +46,9 @@ trait DayAggregateEventController extends BaseController {
   }
 
   private def find(location: Option[Location], skills: Seq[SkillType] = Nil) = {
-    eventsRepository.getEvents(None, None, location, skills).map {
-      _.groupBy(e => DayAggregateEvent(e.date, e.location)).keys.toList
+    eventsRepository.getEvents(None, None, location, skills).map { events =>
+      events.distinctTransform(_.date, e => DayAggregateEvent(e.date, Location("Home"))) ++
```
vglushak: Not verbose. I don't think distinctTransform function needed. It took me a while to understand what is that.  Basically you can do groupBy date and then map to DAE with Home Location.

vglushak: Do we need to put Home value in yaml file? There are places in code where we validate Venues/locations against  available values there.

benjstephenson: Home is a valid location now, and there are places across the code base where the input location is validated against the yaml file so will throw exceptions if it's not there.

vglushak:
I think you should also change `findAvailableAssessorsForLocationAndDate`, otherwise assessors who share their availability for Home won't comes up on a list for event with Location X.

benjstephenson:
👍  I'd missed these changes you mentioned in the last push.

PR #542, opened 2017-09-13
NOJIRA: Update event schedule

henricook:
Update event schedule from a CSV provided by CSHR


fayimora:
`0`s for skills is a no-no 😛

fayimora:
It might be worth updating `VenueType` with `LONDON_FSB`.

PR #544, opened 2017-09-13
FSET-1758: Ensure accepted FastPass candidates are progressed to SIFT

fayimora:


henricook:
@fayimora what happens if the schemes they've selected aren't siftable? (e.g. just choosing human resources)?

fayimora:
@henricook Good catch! My guess is they should get moved straight to `FSAC_AWAITING_ALLOCATION`. I'll confirm with Andy so he can update the ticket.

henricook:
:+1: - i will DNM temporarily

app/services/fastpass/FastPassService.scala
```diff
@@ -82,6 +87,19 @@ trait FastPassService extends EventSink {
...
 
+  private def autoProgressToSiftOrFSAC(applicationId: String): Future[Unit] = {
+    for {
+      preferences <- schemePreferencesService.find(applicationId)
+    } yield {
+      val hasSiftableScheme = schemesRepository.siftableSchemeIds.intersect(preferences.schemes).isEmpty
+      if(hasSiftableScheme) {
+        appRepo.addProgressStatusAndUpdateAppStatus(applicationId, ProgressStatuses.SIFT_ENTERED)
+      } else {
+        appRepo.addProgressStatusAndUpdateAppStatus(applicationId, ProgressStatuses.ASSESSMENT_CENTRE_AWAITING_ALLOCATION)
```
henricook: I suspect this is returning a Future[Future[Unit]] - which means you'll have an unmapped future without a .flatMap(identity)

henricook: The way Future[Unit] eats other futures makes me so sad :-(

fayimora: This is very strange behaviour. I'm very curious why this happens(perhaps some weird type erasure). Thanks for the catch

PR #559, opened 2017-09-25
FSET-1804 changes to support Sdip in 2018FS

ian-stainer:


app/services/onlinetesting/ApplicationStatusCalculator.scala
```diff
@@ -41,11 +41,42 @@ trait ApplicationStatusCalculator {
...
+        case _ => Some(PHASE2_TESTS_PASSED)
+      }
+
+    def phase3SdipFaststreamNotFailed(result: model.EvaluationResults.Result) =
+      result match {
+        case Amber => Some(PHASE3_TESTS_FAILED_SDIP_AMBER)
+        case _ => Some(PHASE3_TESTS_PASSED)
+      }
+
+    val sdipResult = evaluatedSchemes.filter(_.schemeId == SchemeId(sdip)).map(s => Result(s.result)).head
```
mgilgar: From my point of view, arguably it will be more readable, in the sense that you do not have go back and forth to check the functions, the following code:


    val sdipResult = evaluatedSchemes.filter(_.schemeId == SchemeId(sdip)).map(s => Result(s.result)).head
     sdipResult match {
       case Amber if fsOverallResult.contains(PHASE1_TESTS_FAILED) => Some(PHASE1_TESTS_FAILED_SDIP_AMBER)
       case Green if fsOverallResult.contains(PHASE1_TESTS_FAILED) => Some(PHASE1_TESTS_PASSED)
       case Amber if fsOverallResult.contains(PHASE2_TESTS_FAILED) => Some(PHASE2_TESTS_FAILED_SDIP_AMBER)
       case Green if fsOverallResult.contains(PHASE2_TESTS_FAILED) => Some(PHASE2_TESTS_PASSED)
       case Amber if fsOverallResult.contains(PHASE3_TESTS_FAILED) => Some(PHASE3_TESTS_FAILED_SDIP_AMBER)
       case Green if fsOverallResult.contains(PHASE3_TESTS_FAILED) => Some(PHASE3_TESTS_PASSED)
       case Amber if fsOverallResult.contains(PHASE3_TESTS_PASSED) => Some(PHASE3_TESTS_PASSED_WITH_AMBER)
       case Green if fsOverallResult.isEmpty && phase == PHASE3 => Some(PHASE3_TESTS_PASSED_WITH_AMBER)
       case _ => fsOverallResult
     }

otherwise I think phase3SdipFaststreamNotFailed are a bit confusing because they are not verbs, I think is something about converting a result to an status, maybe toResultWhenPhase3SdipFaststreamNotFailed

app/services/onlinetesting/CurrentSchemeStatusHelper.scala
```diff
@@ -0,0 +1,49 @@
...
+  val applicationRepository: GeneralApplicationRepository
+
+  def getSdipResults(application: ApplicationReadyForEvaluation): Future[List[SchemeEvaluationResult]] =
+    if (application.isSdipFaststream) {
+      for {
+        currentSchemeStatus <- applicationRepository.getCurrentSchemeStatus(application.applicationId)
+      } yield {
+        val sdipResultOpt = currentSchemeStatus.find(_.schemeId == SchemeId("Sdip"))
+
+        sdipResultOpt match {
```
mgilgar: I think  you can do sdipResultOpt.toList

app/services/onlinetesting/phase2/EvaluatePhase2ResultService.scala
```diff
@@ -50,6 +52,13 @@ trait EvaluatePhase2ResultService extends EvaluateOnlineTestResultService[Phase2
       case _ => throw new IllegalStateException(s"Illegal number of phase2 active tests with results " +
         s"for this application: ${application.applicationId}")
     }
-    savePassMarkEvaluation(application, schemeResults, passmark)
+
+    CurrentSchemeStatusHelper.getSdipResults(application).map { sdip =>
+      if (application.isSdipFaststream) {
```
mgilgar: Don't you already at this point, the application is sdip faststream?

I think if it is not sdipFastStream, function getSdipResults will return None.

app/services/onlinetesting/phase3/EvaluatePhase3ResultService.scala
```diff
@@ -64,6 +65,13 @@ trait EvaluatePhase3ResultService extends EvaluateOnlineTestResultService[Phase3
       case _ => throw new IllegalStateException(s"Illegal number of phase3 active tests with results " +
         s"for this application: ${application.applicationId}")
     }
-    savePassMarkEvaluation(application, schemeResults, passmark)
+
+    CurrentSchemeStatusHelper.getSdipResults(application).map { sdip =>
+      if (application.isSdipFaststream) {
```
mgilgar: Don't you already at this point, the application is sdip faststream?

I think if it is not sdipFastStream, function getSdipResults will return None.

test/services/onlinetesting/phase1/BakPhase1TestEvaluationSpec.scala
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+// TODO MIGUEL: Remove this
```
mgilgar: I have to remove this class, sorry.

app/services/onlinetesting/phase2/EvaluatePhase2ResultService.scala
ian-stainer: Here we are mapping the future - CurrentSchemeStatusHelper.getSdipResults method returns Future[List[SchemeEvaluationResult]] so if we are not sdip faststream we get back an empty list there is no option here

app/services/onlinetesting/phase3/EvaluatePhase3ResultService.scala
ian-stainer: See comment above, Miguel

app/services/onlinetesting/phase2/EvaluatePhase2ResultService.scala
mgilgar: It feels weird to return a List, when we know is either one or zero.

I can see it is very convenient to have it like a list, there may be a better way but probably not worthy to spend more time.

PR #566, opened 2017-09-29
Map unmapped future from fset-1804

henricook:


mgilgar:
LGTM, I am running the test now.

PR #570, opened 2017-09-29
FSET-1810: Current scheme status should include information about where schemes were first made red

henricook:


app/services/application/ApplicationService.scala
```diff
@@ -298,4 +314,49 @@ trait ApplicationService extends EventSink with CurrentSchemeStatusHelper {
...
+      repository.getPassMarkEvaluation(applicationId).map(_.result).recoverWith { case _ => Future.successful(Nil) }
+
+    for {
+      phase1Evaluation <- getOnlineTestEvaluation(phase1EvaluationRepository)
+      phase2Evaluation <- getOnlineTestEvaluation(phase2EvaluationRepository)
+      phase3Evaluation <- getOnlineTestEvaluation(phase3EvaluationRepository)
+      fsacEvaluation <- fsacRepo.getFsacEvaluatedSchemes(applicationId).map(_.getOrElse(Nil).toList)
+      fsbEvaluation <- fsbRepo.findByApplicationId(applicationId).map(_.map(_.evaluation.result).getOrElse(Nil))
+      evaluations = ListMap(
+        "online tests" -> phase1Evaluation,
```
fayimora: I think we should move these into an enum before it starts getting out of hand. 🤔 `ApplicationPhases` or whatever

henricook: Feels a bit like a premature optimisation at this point, since this is the only place it's required - how would you recommend structuring this enum?

fayimora: Ok @henricook , I'll let this one slide 👀

henricook: You're a nice guy, guy

PR #585, opened 2017-10-13
FSET-1780 handle sdip faststream candidate withdrawal scenarios

ian-stainer:


app/services/application/ApplicationService.scala
```diff
@@ -133,6 +133,7 @@ trait ApplicationService extends EventSink with CurrentSchemeStatusHelper {
       }
   }
 
+  //scalastyle:off
```
henricook: What was scalastyle complaining about here? can you turn off the specific rule e.g. // scalastyle:off method.length

ian-stainer: also complaining about cyclomatic complexity

PR #598, opened 2017-10-20
FSET-1845 Fixed bug to do with retrieving the candidate's latest progress status when working out if they should be progressed to ASSESSMENT_CENTRE_FAILED_SDIP_GREEN_NOTIFIED

ian-stainer:
Fixed bug to do with retrieving the candidate's latest progress status when working out if they should be progressed to ASSESSMENT_CENTRE_FAILED_SDIP_GREEN_NOTIFIED

### DIFF
--- a/app/services/application/FinalOutcomeService.scala
+++ b/app/services/application/FinalOutcomeService.scala
@@ -21,6 +21,7 @@ import connectors.{ CSREmailClient, EmailClient }
 import model.ProgressStatuses.{ ASSESSMENT_CENTRE_FAILED, FSB_FAILED }
 import model.{ ProgressStatuses, SerialUpdateResult }
 import model.command.ApplicationForProgression
+import org.joda.time.DateTime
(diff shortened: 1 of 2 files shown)


app/services/application/FinalOutcomeService.scala
```diff
@@ -78,19 +79,19 @@ trait FinalOutcomeService {
           for {
             (candidate, contactDetails) <- retrieveCandidateDetails(app.applicationId)
             _ <- emailClient.notifyCandidateOnFinalFailure(contactDetails.email, candidate.name)
-            progressStatuses <- applicationRepo.getLatestProgressStatuses
+            progressStatuses <- applicationRepo.getProgressStatusTimestamps(app.applicationId)
```
fayimora: Let's call this progressStatusTimestamps, fetch the last here, extra the status and pass it to progressToNotified.

ian-stainer: I changed the impl. to sort the progress statuses first in date time order and then get the last one.

PR #608, opened 2017-10-26
FSET-1867: Ensure FastPass candidates with siftable schemes receive SIFT_ENTERED email

fayimora:


app/services/fastpass/FastPassService.scala
```diff
@@ -92,14 +91,24 @@ trait FastPassService extends EventSink with CurrentSchemeStatusHelper {
...
+
     val res = for {
       preferences <- schemePreferencesService.find(applicationId)
     } yield {
       val hasSiftableScheme = schemesRepository.siftableSchemeIds.intersect(preferences.schemes).nonEmpty
       if (hasSiftableScheme) {
         val siftStatus = applicationSiftService.progressStatusForSiftStage(preferences.schemes)
-        appRepo.addProgressStatusAndUpdateAppStatus(applicationId, siftStatus)
+        appRepo.addProgressStatusAndUpdateAppStatus(applicationId, siftStatus).flatMap { _ =>
+          notifySiftEntered(siftStatus).map(_ => ())
```
kjharkin: This will seem a subtle point but the 'leaf' method, in this case ```notifySiftEntered``` should not have the if condition. Happy to discuss further.

PR #610, opened 2017-10-27
FSET-1866 Sdip FS candidates who have been sifted out of FS schemes or withdrawn should be invited to FSB if Sdip is sifted in

ian-stainer:
added new state SIFT_FASTSTREAM_FAILED_SDIP_GREEN to identify sdip faststream candidates whose fast stream schemes have either been sifted out or have been withdrawn but whose sdip has been sifted in. These candidates should be invited to attend FSB


kjharkin:
:thumbsup:

PR #611, opened 2017-10-27
FSET-1866 progress candidates straight to FSB who are in SIFT_COMPLETED and withdraw from all fs schemes, leaving just Sdip

ian-stainer:
Also consider candidates who should be progressed to ASSESSMENT_CENTRE_AWAITING_ALLOCATION but have not yet been picked up by the job and so are still in SIFT_COMPLETED when withdrawing from schemes, just leaving Sdip. These candidates should be progressed straight to FSB


fayimora:
Good catch! 👍

PR #616, opened 2017-10-30
FSET-1853: Add or replace an analysis exercise file for a candidate

fayimora:


app/controllers/ApplicationController.scala
```diff
@@ -200,6 +216,20 @@ trait ApplicationController extends BaseController {
       }
   }
 
+  def retrieveAnalysisExerciseInfo(applicationId: String): Action[AnyContent] = Action.async {
```
henricook: I was confused when reading this because of "info" and "meta data", they both sounded like the same thing - could you rename this endpoint to perhaps retrieveAnalysisExerciseId?

fayimora: `retrieveAnalysisExerciseId` implies that all I'm retrieving is the file id but it could have more attributes right?

henricook: I see that assessmentCentreTests returns an object. The two endpoints are still confusingly named, what's a better solution?

Call the second one analysisExerciseFileMetadata perhaps? Also pedantry shoutout, metadata is one work so no capital D in camel case :-)

fayimora: Renamed the other function

PR #619, opened 2017-10-31
FSET-1875 new end points to fix candidates stuck in SIFT_ENTERED

ian-stainer:
added end points to find and fix candidates who are stuck in SIFT_ENTERED who need to be moved to SIFT_READY because they are no longer in the running for form based schemes so need to just be sifted


app/repositories/sift/ApplicationSiftRepository.scala
```diff
@@ -247,5 +247,26 @@ class ApplicationSiftMongoRepository(
       )
     })
   }
+
+  def findAllUsersInSiftEntered: Future[Seq[FixUserStuckInSiftEntered]] = {
+
+    val query = BSONDocument("applicationStatus" -> ApplicationStatus.SIFT,
```
henricook: I think it's possible to skip SIFT_READY, e.g. if you have only PDFS (Form required: true, evaluation required: false) - so in that case this would still return true. I'd perhaps add a _SIFT_COMPLETE -> exists false_ as well

PR #630, opened 2017-11-09
FSET-1886: Increase batch size of FSAC evaluations

fayimora:


app/scheduler/assessment/EvaluateAssessmentScoreJob.scala
```diff
@@ -36,24 +36,26 @@ trait EvaluateAssessmentScoreJob extends SingleInstanceScheduledJob[BasicJobConf
...
     }
   }
 }
 
 object EvaluateAssessmentScoreJobConfig extends BasicJobConfig[WaitingScheduledJobConfig](
   configPrefix = "scheduling.evaluate-assessment-score-job",
   name = "EvaluateAssessmentScoreJob"
-)
+) {
+  val batchSize: Int = conf.batchSize.getOrElse(throw new IllegalArgumentException("Batch size must be defined"))
```
henricook: Lets default this to 1 instead of throwing an exception, i think we do this elsewhere

PR #642, opened 2017-11-22
FSET-1906: Add endpoint to mark a scheme's sift result as red

henricook:


app/repositories/sift/ApplicationSiftRepository.scala
```diff
@@ -301,5 +303,53 @@ class ApplicationSiftMongoRepository(
...
     }
   }
+
+  def fixDataByRemovingSiftEvaluation(applicationId: String): Future[Unit] = {
+    val query = BSONDocument("applicationId" -> applicationId)
+
+    val updateOp = bsonCollection.updateModifier(
+      BSONDocument(
+        "$set" -> BSONDocument("applicationStatus" -> ApplicationStatus.SIFT),
+        "$unset" -> BSONDocument(s"testGroups.$phaseName" -> "")
```
fayimora: @henricook this un-sets the entire sift phase. I think we want to remove the result for a particular scheme here.

henricook: The ticket hasn't specified a particular scheme, and i believe we talked about removing their whole evaluation so they could just be re-sifted

henricook: The ticket contains two candidates, one who needs an individual scheme red'ing and one who needs to be rolled back from AC_AWAITING_ALLOCATION

henricook: The currentSchemeStatus also needs resetting, i've had some crossover here so took care of it in the PR for FSET-1909

PR #643, opened 2017-11-22
FSET-1901 & FSET-1902: SdipFaststream candidates should fail all FS schemes and progress to SIFT when online tests expire and SDIP is GREEN

fayimora:


henricook:
I'd quite like to see how the manual fix for this plays out before rolling this in, additionally with almost no one left to go through video interview this year i don't think we need to rush this in.

PR #650, opened 2017-12-01
FSET-1915: Fix issue with failed_at_sift rollback by manually tweaking currentschemestatus

henricook:


app/services/application/ApplicationService.scala
```diff
@@ -513,6 +513,20 @@ trait ApplicationService extends EventSink with CurrentSchemeStatusHelper {
...
+  def updateCurrentSchemeStatusScheme(applicationId: String, schemeId: SchemeId, newResult: model.EvaluationResults.Result): Future[Unit] = {
+    for {
+      currentSchemeStatus <- appRepository.getCurrentSchemeStatus(applicationId)
+      newCurrentSchemeStatus = currentSchemeStatus.map { schemeResult =>
+        if (schemeResult.schemeId == schemeId) {
+          schemeResult.copy(result = newResult.toString)
+        } else {
+          schemeResult
+        }
+      }
```
fayimora: This will complete silently if there is no scheme to be updated. Probably not required but maybe we want to call `updateCurrentSchemeStatus ` only if we have an update to perform?

```scala
for {
  currentSchemeStatus <- appRepository.getCurrentSchemeStatus(applicationId)
  schemeExists = currentSchemeStatus.exists(_.schemeId == schemeId) if schemeExists
  newCurrentSchemeStatus = currentSchemeStatus.map { ..... }
  _ <- appRepository.updateCurrentSchemeStatus(applicationId, newCurrentSchemeStatus)
}
```

OR perhaps just log the operation? OR both

PR #656, opened 2017-12-15
FSET-1930: Ensure optimistic locking for sifting is done per scheme

fayimora:


fayimora:
Tests are all passing locally so will merge.

PR #680, opened 2018-02-12
FSET-1975: In progress GES-DS candidates (two FSBs required) cause exceptions to be thrown to the error log

henricook:


ian-stainer:
LGTM - so the team can discuss if we want to include this before merging

PR #681, opened 2018-02-14
FSET-1935 Use renamed cubiks gateway

ian-stainer:
Changed the CubiksGatewayClient to use the renamed cubiks gateway (fset-online-tests-gateway). Also added an endpoint to manually trigger the invitations to the online tests for each phase


henricook:
Needs an application.conf change paired with it to point to the fset-online-tests-gateway locally, and needs app-config-* changes for environments

PR #686, opened 2018-03-16
FSET-2014 added method to EmailClient so we can send out the sift expired email

ian-stainer:


fayimora:
@ian-stainer We'll need something in the app service(or somewhere appropriate) that takes in an `appId`(or multiple) and fetches personal details before calling the email client.

PR #693, opened 2018-04-10
FSET-2004 added endpoint to allow us to extend the sift phase

ian-stainer:


app/model/persisted/sift/SiftTestGroup.scala
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package model.persisted.sift
+
+import org.joda.time.DateTime
+import play.api.libs.json.Json
+import reactivemongo.bson.{ BSONDocument, BSONHandler, Macros }
+
+case class SiftTestGroup(expirationDate: DateTime)
```
fayimora: If this represents the entire sift test group, then it should also have the `evaluation` field(perhaps as `Option[T]`). This way, it can be re-used at later stages and for reports.

app/services/sift/SiftExpiryExtensionService.scala
```diff
@@ -0,0 +1,110 @@
...
+  }
+}
+
+case class SiftExtensionException(message: String) extends Exception(message)
+
+private final case class Extension(extendedExpiryDate: DateTime, expired: Boolean, progress: ProgressResponse)
+
+object SiftExpiryExtensionServiceImpl {
+
+  val NoOp: Future[Unit] = Future.successful(())
```
fayimora: 👍

PR #730, opened 2018-06-27
update fset-faststream with HTTP to https

emailvacharya:


ian-stainer:
Not needed. So will close.

PR #742, opened 2018-07-17
FSET-2088 change the collection name suffix to 19test in preparation for the new campaign setup

ian-stainer:


emailvacharya:
LGTM

PR #906, opened 2019-07-15
Fset 2375

mgilgar:


sbtTestRoutes.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-sbt -J-Dapplication.router=testOnlyDoNotUseInAppConf.Routes -Dhttp.port=8101
+sbt -jvm-debug 7101 -J-Dapplication.router=testOnlyDoNotUseInAppConf.Routes -Dhttp.port=8101
```
fayimora: You don't want to commit this 😊

PR #917, opened 2019-07-24
FSET-2372: Add ability to reset PHASE1 tests

fayimora:


app/services/onlinetesting/phase1/Phase1TestService2.scala
```diff
@@ -122,9 +122,58 @@ trait Phase1TestService2 extends OnlineTestService with Phase1TestConcern2 with
...
+      testGroup = maybeTestGroup
+        .getOrElse(throw CannotFindTestGroupByApplicationId(s"appId - ${application.applicationId}"))
+
+      // Extract test that requires reset
+      testToReset = testGroup.tests.find(_.orderId == orderIdToReset)
+        .getOrElse(throw CannotFindTestByOrderId(s"OrderId - $orderIdToReset"))
+      _ = Logger.info(s"testToReset -- $testToReset")
+
+      // Create PsiIds to use for re-invitation
+      psiIds = integrationGatewayConfig.phase1Tests.tests.find {
```
mgilgar: Is integrationGatewayConfig a global variable, how are we going to unit test it?

app/services/onlinetesting/phase1/Phase1TestService2.scala
```diff
@@ -122,9 +122,58 @@ trait Phase1TestService2 extends OnlineTestService with Phase1TestConcern2 with
...
+        case (_, ids) => ids.inventoryId == testToReset.inventoryId
+      }.getOrElse(throw CannotFindTestByInventoryId(s"InventoryId - ${testToReset.inventoryId}"))._2
+      _ = Logger.info(s"psiIds -- $psiIds")
+
+      // Register applicant
+      newPsiTest <- registerPsiApplicant(application, psiIds, invitationDate)
+      _ = Logger.info(s"newPsiTest -- $newPsiTest")
+
+      // Set old test to inactive
```
mgilgar: Shouldn't we make it inactive before we create the new test?

fayimora: @mgilgar 
We could but I wanted to set it to inactive only after the test creation was successful. That way, we only change the test when we have 100% certainty that there's something to "replace" it with.

PR #992, opened 2020-02-05
FSET-2504 migrating from play-reactivemongo to simple-reactivemongo

ian-stainer:


ian-stainer:
I have created a replacement pr, which includes a rebase from master and all the corresponding commits have now been GPG signed. That supersedes this one so this pr is redundant

